PR #2, opened 2025-03-30
refactor: change to factory pattern, and add luxos discovery

b-rowan:
Updated to use a factory pattern, while leaving the simple `get_miner` call at `asic_rs::get_miner`.  Also added luxos model discovery support.

Noticed a minor issue with this way of handling models, for firmware based identification we don't yet have a good way to discriminate between makes based on firmware.  EG BraiinsOS on Antminers vs BraiinsOS on Braiins Mini Miner isn't really covered by this handler.


src/miners/factory/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        let miner_info = tokio::select!(
+            Ok(miner_info) = &mut tasks => {
+                miner_info
+            },
+            _ = &mut timeout => {
+                None
+            }
+        );
+
+        match miner_info {
```
jpcomps, line 173: i think we may be able to match the combinations here for situations like Antminer, BOS or BOS,BOS

jpcomps:
looks good, lgtm

src/miners/factory/mod.rs
b-rowan: Yeah, then I think we end up having to make a `make` decision during initial discovery, when AFAIU we still don't have enough information to make that distinction.  EG BraiinsOS we may be able to choose using the RPC result (haven't checked, but I assume it could be refactored to do that if needed), but the web result will not indicate it's `make`.  We have to decide on the `make` as late as possible...

jpcomps: Ahh I see, yeah in that case may just to figure out the FW first, then from there determine the make. I "think" we should always be able to determine the FW type, and each one should be able to tell us the model. So its a blocking in the sense we will need FW first, but then we can always guarantee to figure out the model. 

Experimentation needed for sure. Other way is later down the line we just swap in the proper model using something like an ArcSwap (don't think its needed yet)

b-rowan: It is also possible that we could just make `MinerModel.from_str()` make agnostic and just return the proper type, then in data rather than having the `make` and `model` we just use `model`, since it would encode both.

PR #5, opened 2025-04-26
feature: add RPC API design for LUXMiner

b-rowan:
Attempt a setup for an RPC API (for LUXMiner since that's what I have access to right now).  Couple caveats here  with the way this is being done, to try to add some flexibility in the future:

- The `SendRPCCommand` trait expects a type `T` which in the general case will be `serde_json::Value`, which can be a bit annoying, but the goal with this is that ideally the user should not use this function, and instead internally we can optimize memory usage by parsing the raw socket response directly into a type that contains all the data we want, then discarding the rest of the response.  We would implement a type `LUXMinerDevdetailsData` which would parse only the important data, which could then be put into `MinerData`.  This also solves the same problem as the dynamic `_get_data` function in pyasic in that it will ensure that each unique command is only ever sent 1 time.
- `send_rpc_command` does not take a `MinerCommand` (yet), as I think that type will need to be extended with parameters.  Not sure how best to handle this, so leaving this starting implementation as basic as possible to review the underlying idea.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,11 +1,16 @@
+use asic_rs::miners::api::rpc::luxminer::LUXMinerRPC;
+use asic_rs::miners::api::rpc::traits::SendRPCCommand;
+use serde_json::Value;
 use std::net::IpAddr;
-
-use asic_rs::get_miner;
 
 #[tokio::main]
 async fn main() {
     let miner_ip = IpAddr::from([192, 168, 86, 34]);
+    let port: u16 = 4028;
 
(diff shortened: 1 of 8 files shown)


jpcomps:
I'm good with the concept. Might be a way to make it fully generic and take in two types (Protocol, MinerType). Will have a think on that but this seems like the right way to go. Apologies for the delay on some of this arch work. Gonna also tag @cfilipescu for some advice. Hopefully in the next couple weeks we will get a chance to dedicate some time to this

PR #11, opened 2025-07-25
Update DataFields and ESPMiner

lurkny:
Made the DataField enum match what we expect in MinerData, and updated the espminer backend accordingly.

### DIFF
--- a/src/data/device/models/bitaxe.rs
+++ b/src/data/device/models/bitaxe.rs
@@ -1,9 +1,12 @@
 use serde::Serialize;
-
 #[derive(Debug, Clone, PartialEq, Eq, Serialize)]
 pub enum BitaxeModel {
+    #[serde(alias = "BM1368")]
     Supra,
+    #[serde(alias = "BM1370")]
     Gamma,
+    #[serde(alias = "BM1397")]
     Max,
+    #[serde(alias = "BM1366")]
     Ultra,
 }
--- a/src/data/device/models/bitaxe.rs
+++ b/src/data/device/models/bitaxe.rs
@@ -1,0 +1,9 @@
+use serde::Serialize;
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub enum BitaxeModel {
+    Supra,
+    Gamma,
+    Max,
+    Ultra,
+}
--- a/src/data/device/models/mod.rs
+++ b/src/data/device/models/mod.rs
@@ -1,5 +1,5 @@
 use super::{MinerFirmware, MinerMake};
-use crate::data::device::models::bitaxe::BitaxeModel;
+use bitaxe::BitaxeModel;
 use antminer::AntMinerModel;
 use braiins::BraiinsModel;
 use serde::Serialize;
--- a/src/data/device/models/mod.rs
+++ b/src/data/device/models/mod.rs
@@ -1,4 +1,5 @@
 use super::{MinerFirmware, MinerMake};
+pub(crate) use crate::data::device::models::bitaxe::BitaxeModel;
 use antminer::AntMinerModel;
 use braiins::BraiinsModel;
 use serde::Serialize;
@@ -6,6 +7,7 @@ use std::{fmt::Display, str::FromStr};
 use whatsminer::WhatsMinerModel;
 
 pub mod antminer;
+mod bitaxe;
 pub mod braiins;
 pub mod whatsminer;
 
@@ -52,13 +54,6 @@ pub enum MinerModel {
     WhatsMiner(WhatsMinerModel),
     Braiins(BraiinsModel),
     Bitaxe(BitaxeModel),
-}
-#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
-pub enum BitaxeModel {
-    Supra,
-    Gamma,
-    Max,
-    Ultra,
 }
 
 pub(crate) struct MinerModelFactory {
--- a/src/data/device/models/mod.rs
+++ b/src/data/device/models/mod.rs
@@ -1,5 +1,5 @@
 use super::{MinerFirmware, MinerMake};
-pub(crate) use crate::data::device::models::bitaxe::BitaxeModel;
+use crate::data::device::models::bitaxe::BitaxeModel;
 use antminer::AntMinerModel;
 use braiins::BraiinsModel;
 use serde::Serialize;
@@ -7,7 +7,7 @@ use std::{fmt::Display, str::FromStr};
 use whatsminer::WhatsMinerModel;
 
 pub mod antminer;
-mod bitaxe;
+pub mod bitaxe;
 pub mod braiins;
 pub mod whatsminer;
 
--- a/src/miners/backends/espminer.rs
+++ b/src/miners/backends/espminer.rs
@@ -13,6 +13,7 @@ use crate::data::device::{DeviceInfo, HashAlgorithm, MinerHardware, MinerModel};
 use crate::data::fan::FanData;
 use crate::data::hashrate::{HashRate, HashRateUnit};
 use crate::data::miner::MinerData;
+use crate::data::pool::{PoolData, PoolScheme, PoolURL};
 use crate::miners::api::web::esp_web_api::EspWebApi;
 use crate::miners::backends::traits::GetMinerData;
 use crate::miners::data::{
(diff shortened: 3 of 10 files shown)


lurkny:
Need to hammer out some details, get the hashboards + pool stuff working before PR is ready

src/miners/backends/espminer.rs
```diff
@@ -80,12 +115,49 @@ impl GetMinerData for ESPMiner {
...
             .unwrap_or_default();
 
+        let uptime = data
+            .get(&DataField::Uptime)
+            .and_then(|v| v.as_u64())
+            .map(Duration::from_secs);
+
+        let is_mining = hashrate.as_ref().map_or(false, |hr| hr.value > 0.0);
+
+        // Construct hashboards to match Python, assuming BoardData struct definition
```
b-rowan, line 136: BoardData struct should exist in data::boards

src/miners/backends/espminer.rs
```diff
@@ -129,25 +202,48 @@ impl GetMinerData for ESPMiner {
...
-            DataField::Hostname => &[(
+            DataField::ControlBoardVersion => &[(
                 SYSTEM_INFO_CMD,
                 DataExtractor {
                     func: get_by_key,
-                    key: Some("hostname"),
+                    key: Some("boardVersion"),
+                },
+            )],
+            DataField::ExpectedHashboards => &[
```
b-rowan, line 226: This is a constant per miner, in this case it's 1 board.

src/miners/backends/espminer.rs
```diff
@@ -157,24 +253,43 @@ impl GetMinerData for ESPMiner {
                     key: Some("hashRate"),
                 },
             )],
-            DataField::ExpectedHashrate => &[],
-            DataField::Hashboards => &[],
+            DataField::TotalChips => &[(
```
b-rowan, line 256: This should be a sum of the chips on each board, not cores.

src/miners/backends/espminer.rs
```diff
@@ -157,24 +253,43 @@ impl GetMinerData for ESPMiner {
...
+                SYSTEM_INFO_CMD,
+                DataExtractor {
+                    func: get_by_key,
+                    key: Some("smallCoreCount"),
+                },
+            )],
+            DataField::ExpectedFans => &[],
```
b-rowan, line 263: Fixed per miner, 1 in this case.

b-rowan (changes requested):
Reviewing on my phone, so may not have caught everything, but mostly just issues with my documenting the data structure it seems like.  I'll have to go through and make sure everything is labeled and explained in the MinerData struct.

src/data/device/models/bitaxe.rs
```diff
@@ -0,0 +1,9 @@
+use serde::Serialize;
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub enum BitaxeModel {
+    Supra,
+    Gamma,
+    Max,
+    Ultra,
```
b-rowan, line 8: These should be aliased with serde so the identification function can just pass in the chip type.

```rust
    #[serde(alias = "BM1368")]
````

Easiest way right now will be to just match it to the pyasic factory keys - https://github.com/UpstreamData/pyasic/blob/90c89869009c108e3e85cb3fda42421d5d655d8d/pyasic/miners/factory.py#L667-L673

File: src/data/device/models/mod.rs [hunk 6]
```diff
@@ -1,11 +1,13 @@
 use super::{MinerFirmware, MinerMake};
+pub(crate) use crate::data::device::models::bitaxe::BitaxeModel;
 use antminer::AntMinerModel;
 use braiins::BraiinsModel;
 use serde::Serialize;
 use std::{fmt::Display, str::FromStr};
 use whatsminer::WhatsMinerModel;
 
 pub mod antminer;
+mod bitaxe;
```
b-rowan, line 10: Should be consistent with the rest, I like `pub mod` here.

src/data/device/models/mod.rs
```diff
@@ -1,11 +1,13 @@
 use super::{MinerFirmware, MinerMake};
+pub(crate) use crate::data::device::models::bitaxe::BitaxeModel;
```
b-rowan, line 2: Should be consistent with the rest, probably just `use bitaxe::BitaxeModel`.

src/miners/backends/espminer.rs
```diff
@@ -50,76 +56,237 @@ impl GetMinerData for ESPMiner {
...
+
+        // Get hardware specifications based on the miner model
+        let miner_hardware = MinerHardware::from(&self.model);
+
+        let hashboards = {
+            let board_voltage = data.get(&DataField::Hashboards).and_then(|hashboard_data| {
+                hashboard_data.get("voltage")
+                    .and_then(|voltage_value| voltage_value.as_f64())
+                    .map(Voltage::from_millivolts)
+            });
```
b-rowan, line 133: Formatting on this is bugging me, since the rest are on new lines.  I assume this is a `cargo fmt` issue?  In any case I would like to make this consistent if possible.

src/miners/backends/espminer.rs
```diff
@@ -50,76 +56,237 @@ impl GetMinerData for ESPMiner {
...
+                frequency: board_frequency,
+                tuned: None,
+                working: Some(true),
+                hashrate: board_hashrate.clone(),
+            };
+
+            let board_data = BoardData {
+                position: 0,
+                hashrate: board_hashrate,
+                expected_hashrate: None,
```
b-rowan, line 168: This actually should be in the current version of the API.

See https://github.com/bitaxeorg/ESP-Miner/pull/603

src/miners/backends/espminer.rs
```diff
@@ -50,76 +56,237 @@ impl GetMinerData for ESPMiner {
...
+                board_temperature,
+                intake_temperature: None,
+                outlet_temperature: None,
```
b-rowan, line 171: I'm fine with using board temperature for these for the bitaxes, just because of how they are built.

src/miners/backends/espminer.rs
```diff
@@ -50,76 +56,237 @@ impl GetMinerData for ESPMiner {
...
+                expected_chips: miner_hardware.chips,
+                working_chips: total_chips,
+                serial_number: None,
+                chips: vec![chip_info],
+                voltage: board_voltage,
+                frequency: board_frequency,
+                tuned: None,
```
b-rowan, line 178: Tempted to have this default to `true`, at least for miner types that don't implicitly support some kind of tuning, as we can assume it was tuned by the factory in a roundabout way.  Let me know your thoughts here.

src/miners/backends/espminer.rs
```diff
@@ -50,76 +56,237 @@ impl GetMinerData for ESPMiner {
...
+                port: main_port as u16,
+                pubkey: None,
+            };
+            let main_pool_data = PoolData {
+                position: Some(0),
+                url: Some(main_pool_url),
+                accepted_shares: accepted_share,
+                rejected_shares: rejected_share,
+                active: Some(!is_using_fallback),
+                alive: None,
```
b-rowan, line 211: Sucks we have no way of figuring this out.  Maybe something I have to bring up in espminer.

src/miners/backends/espminer.rs
```diff
@@ -50,76 +56,237 @@ impl GetMinerData for ESPMiner {
...
+                user: None,
```
b-rowan, line 212: This should be the `stratumUser` key.

```rust
let user = data.get(&DataField::Pools).and_then(|r| r.get("stratumUser")).and_then(|s| s.as_str()).unwrap_or("").to_string();
```

src/miners/backends/espminer.rs
```diff
@@ -50,76 +56,237 @@ impl GetMinerData for ESPMiner {
...
+                pubkey: None,
+            };
+            let fallback_pool_data = PoolData {
+                position: Some(1),
+                url: Some(fallback_pool_url),
+                accepted_shares: accepted_share,
+                rejected_shares: rejected_share,
+                active: Some(is_using_fallback),
+                alive: None,
+                user: None,
```
b-rowan, line 231: This is `fallbackStratumUser`.

```rust
let fallback_user = data.get(&DataField::Pools).and_then(|r| r.get("fallbackStratumUser")).and_then(|s| s.as_str()).unwrap_or("").to_string();
```

src/miners/backends/espminer.rs
```diff
@@ -50,76 +56,237 @@ impl GetMinerData for ESPMiner {
...
+            vec![main_pool_data, fallback_pool_data]
+        };
+
         MinerData {
+            // Version information
             schema_version: env!("CARGO_PKG_VERSION").to_string(),
             timestamp,
+            
+            // Network identification
             ip: self.web.ip.clone().parse().unwrap(),
```
b-rowan, line 243: Does `self.ip` exist?  I feel like this should us that...

src/miners/backends/espminer.rs
```diff
@@ -50,76 +56,237 @@ impl GetMinerData for ESPMiner {
...
             mac,
+            
+            // Device identification
             device_info: DeviceInfo::new(BitAxe, self.model.clone(), Stock, HashAlgorithm::SHA256),
```
b-rowan, line 247: Less important for bitaxes, but should we be tracking firmware as part of the top-level struct?  Would be good to have `self.firmware`...

src/miners/backends/espminer.rs
```diff
@@ -50,76 +56,237 @@ impl GetMinerData for ESPMiner {
...
             fluid_temperature: None,
+            
+            // Power information
             wattage,
             wattage_limit: None,
             efficiency,
+            
+            // Status information
             light_flashing: None,
             messages: vec![],
```
b-rowan, line 279: Might be a good idea to construct our own message from this with `overheat_mode` or something.  Will ask the espminer guys about error/message handling.

b-rowan (changes requested):
Couple small things.  Some smaller fixes, then a couple things I'm curious for comment on.

PR #12, opened 2025-07-25
docs: add documentation for `MinerData`

b-rowan:
Added documentation for the `MinerData` struct and subitems.

Should look at adding `::new()` functions for these, as sections of this data should be computed from data at lower levels, such as averages and chip counts.


lurkny (approved):
Looks great!

PR #13, opened 2025-07-28
feature: separate espminer versions in 2.9.0+ and 2.0.0+

b-rowan:


lurkny (changes requested):
Looks great, I just have a couple of suggestions

PR #18, opened 2025-07-31
refacor: restructure backend folders

b-rowan:
Restructured backends folders to make a bit more sense with how this project is composed.  Excluding LUXMiner for now, since it has no `GetMinerData` impl yet.


lurkny (approved):
Looks good!

PR #21, opened 2025-08-01
tests: add espminer v2.0.0 tests and separate `get_data` and `parse_data`

b-rowan:
Adds a basic data checking test for `ESPMiner` v2, and splits out `get_data` and `parse_data` so that custom data can be passed in.


lurkny (approved):
Looks good, only comment I have is to consider moving the mock json responses into their own file and importing them for cleanliness’ sake.

PR #22, opened 2025-08-01
feat(factory): Add subnet scanning support

lurkny:
Added in the ability to scan via subnet.

Also added in some cargo options to optimize our release build, reducing binary size by 66%.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,6 @@ dependencies = [
  "anyhow",
  "async-trait",
  "diqwest",
- "dyn_serde",
  "futures",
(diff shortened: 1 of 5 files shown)


b-rowan:
@s0kil this is why my thought was just to use the `MinerFactory`, as we can use `with_retries`.  Then the main focus is on making the factory scan very efficient, which it already is because of the way requests are grouped.

The downside of the scan you have as I see it is that it hits the dead hosts multiple times, and guarantees that you are waiting the maximum amount of time on every scan.

b-rowan (approved):

PR #23, opened 2025-08-05
feat: Add Support for Avalon Miners

lurkny:


b-rowan (changes requested):
Lots of stuff to thing about here.  Very difficult miner to implement, but need to separate some stuff into traits and generally try to deal with clutter.  Great start in any case.

src/miners/backends/traits.rs
```diff
@@ -455,6 +455,36 @@ pub trait GetLightFlashing: CollectData {
...
+pub trait Pause {
+    async fn pause(&self, at_time: Option<u64>) -> Result<()>;
+
+}
+
+#[async_trait]
+pub trait Resume {
+    async fn resume(&self, at_time: Option<u64>) -> Result<()>;
+}
+
```
b-rowan: I feel like most of these traits should return `Result<bool>`, since I want to use `Err` to represent things like connection errors.  This will help later on when, for example, an external application is trying to control the machines on behalf of a user.  An `Err` will tell them something went wrong with the machine, such as a connection error, and they may need to rescan as it may have moved, but a `false` value will tell them that maybe this miner doesn't support that.

b-rowan (changes requested):

s0kil:
@lurkny Have a few avalons running at our site, willing to help test this pr, ping me when ready

PR #32, opened 2025-08-07
Initial VnishOS Backend

s0kil:


meta/vnish-api-doc.json
b-rowan, line 1: Can we version this somehow?  Chances are they change this in the future, so maybe this can be at `meta/vnish/v1.2.6/apenapi.json` or something?

src/data/device/models/antminer.rs
```diff
@@ -3,107 +3,166 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, PartialEq, Eq, Clone, Copy, Serialize, Deserialize)]
 pub enum AntMinerModel {
     #[serde(alias = "ANTMINER D3")]
+    #[serde(alias = "d3")]
```
b-rowan, line 6: Would like to uppercase this so that if and when they change it we don't have to deal with a breaking change in the factory.

src/miners/backends/vnish/mod.rs
```diff
@@ -0,0 +1,469 @@
...
+
+#[derive(Debug)]
+pub struct Vnish {
+    ip: IpAddr,
+    web: VnishWebAPI,
+    device_info: DeviceInfo,
+}
+
+impl Vnish {
+    pub fn new(ip: IpAddr, model: MinerModel, firmware: MinerFirmware) -> Self {
```
b-rowan, line 33: This should take in `model` and `make` or something, we know what the firmware is already.

Eventually I will fix up the PR for unifying the make and model then this will be a non-issue.

src/miners/backends/vnish/mod.rs
```diff
@@ -0,0 +1,469 @@
...
+                            .and_then(|v| v.as_u64())
+                            .unwrap_or(idx as u64) as u8,
+                        hashrate,
+                        expected_hashrate,
+                        board_temperature,
+                        intake_temperature: chip_temperature,
+                        outlet_temperature: chip_temperature,
+                        expected_chips: self.device_info.hardware.chips,
+                        working_chips,
+                        serial_number: None, // Not provided in AntmChainChips schema
```
b-rowan, line 322: Should be able to get this from `/chains/factory-info` AFAIU.

src/miners/backends/vnish/mod.rs
```diff
@@ -0,0 +1,469 @@
...
+            )],
+            DataField::Hashrate => vec![(
+                summary_cmd.clone(),
+                DataExtractor {
+                    func: get_by_pointer,
+                    key: Some("/miner/hr_realtime"),
+                },
+            )],
+            DataField::ExpectedHashrate => vec![(
+                summary_cmd.clone(),
```
b-rowan, line 124: What?  Why?

src/miners/backends/vnish/mod.rs
```diff
@@ -0,0 +1,469 @@
...
+            )],
+            DataField::Fans => vec![(
+                summary_cmd.clone(),
+                DataExtractor {
+                    func: get_by_pointer,
+                    key: Some("/miner/cooling/fans"),
+                },
+            )],
+            DataField::Hashboards => vec![(
+                chains_cmd,
```
b-rowan, line 145: May want to look into using `/summary` for this, it provides a LOT of information.

Example value:
```
{
  "miner": {
    "best_share": 0,
    "chains": [
      {
        "chip_statuses": {
          "grey": 0,
          "orange": 0,
          "red": 0
        },
        "chip_temp": {
          "max": 0,
          "min": 0
        },
        "frequency": 0.1,
        "hashrate_ideal": 0.1,
        "hashrate_percentage": 0.1,
        "hashrate_rt": 0.1,
        "hr_error": 0.1,
        "hw_errors": 0,
        "id": 0,
        "inlet_water_temp": 0,
        "outlet_water_temp": 0,
        "pcb_temp": {
          "max": 0,
          "min": 0
        },
        "power_consumption": 0,
        "status": {
          "description": "string",
          "state": "initializing"
        },
        "voltage": 0
      }
    ],
    "chip_temp": {
      "max": 0,
      "min": 0
    },
    "cooling": {
      "fan_duty": 0,
      "fan_num": 0,
      "fans": [
        {
          "id": 0,
          "max_rpm": 0,
          "rpm": 0,
          "status": "ok"
        }
      ],
      "settings": {
        "mode": {
          "name": "manual"
        }
      }
    },
    "devfee": 0.1,
    "devfee_percent": 0.1,
    "found_blocks": 0,
    "hr_average": 0.1,
    "hr_error": 0.1,
    "hr_nominal": 0.1,
    "hr_realtime": 0.1,
    "hr_stock": 0.1,
    "hw_errors": 0,
    "hw_errors_percent": 0.1,
    "miner_status": {
      "description": "string",
      "failure_code": 0,
      "miner_state": "mining",
      "miner_state_time": 0
    },
    "miner_type": "string",
    "pcb_temp": {
      "max": 0,
      "min": 0
    },
    "pools": [
      {
        "accepted": 0,
        "asic_boost": true,
        "diff": "string",
        "diffa": 0.1,
        "id": 0,
        "ls_diff": 0.1,
        "ls_time": "string",
        "ping": 0,
        "pool_type": "UserPool",
        "rejected": 0,
        "stale": 0,
        "status": "offline",
        "url": "string",
        "user": "string"
      }
    ],
    "power_consumption": 0,
    "power_efficiency": 0.1,
    "psu": {
      "psu_power_metering": true,
      "temps": null
    }
  }
}
```

It is called `AntmMinerStats` in the schema.

src/miners/backends/vnish/mod.rs
```diff
@@ -0,0 +1,469 @@
...
+                for (idx, pool) in pools_array.iter().enumerate() {
+                    let url = pool
+                        .pointer("/url")
+                        .and_then(|v| v.as_str())
+                        .map(|url_str| {
+                            // Parse the URL - assume stratum format
+                            PoolURL {
+                                scheme: PoolScheme::StratumV1,
+                                host: url_str.to_string(),
+                                port: 4444, // Default stratum port
```
b-rowan, line 433: Nope?  I think we can use `PoolURL::from()` for this, it isnt clear how the URL is formatted but I hope it has the port.

src/miners/backends/vnish/web.rs
```diff
@@ -0,0 +1,221 @@
...
+    pub fn with_api_key(mut self, api_key: String) -> Self {
+        self.api_key = Some(api_key);
+        self
+    }
+
+    /// Set bearer token for authentication (Authorization: Bearer header)
+    pub fn with_bearer_token(mut self, token: String) -> Self {
+        self.bearer_token = Some(token);
+        self
+    }
```
b-rowan, line 136: These methods are not part of the `GetMinerData` trait, so they are essentially completely unusable.  We could maybe generalize these, but I personally am fine with just getting rid of them for now and adding them if we need to.  Might also be an option to set a global setting.

src/miners/backends/vnish/web.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        let response = self
+            .client
+            .execute(request)
+            .await
+            .map_err(|e| VnishError::NetworkError(e.to_string()))?;
+
+        Ok(response)
+    }
+}
```
b-rowan, line 186: Need to automate authentication here, I think there needs to be a plan made as to how we are going to handle login information for machines.  Obviously something we want on the API, but needs to be generalized so that the top level struct (`dyn GetMinerData`) can implement methods for it.

See pyasic implementation of this function here - https://github.com/UpstreamData/pyasic/blob/b8ae238d2337e106431a3a681f4c0ae63e7d89a9/pyasic/web/vnish.py#L28-L94

src/miners/factory/model/mod.rs
```diff
@@ -9,6 +9,51 @@ use std::net::IpAddr;
 
 pub mod whatsminer;
 
+pub(crate) async fn get_model_vnish(ip: IpAddr) -> Option<MinerModel> {
+    let response: Option<Response> = Client::new()
+        .get(format!("http://{}/api/v1/info", ip))
```
b-rowan, line 14: Do these need authentication?  I can never remember.  VNish still supports the standard RPC API, so we may want to use that here.  In any case, should make use of `util::send_rpc_command` or `util::send_web_command`.

pyasic's factory implementation of this - https://github.com/UpstreamData/pyasic/blob/b8ae238d2337e106431a3a681f4c0ae63e7d89a9/pyasic/miners/factory.py#L1253-L1269

src/miners/factory/mod.rs
```diff
@@ -134,6 +135,7 @@ fn select_backend(
         (Some(MinerMake::BitAxe), Some(MinerFirmware::Stock)) => {
             Some(ESPMiner::new(ip, model?, firmware?, version?))
         }
+        (_, Some(MinerFirmware::VNish)) => Some(Box::new(Vnish::new(ip, model?, firmware?))),
```
b-rowan, line 138: Am liking the idea of versioning this, so something we may want to think about.  Nice cause once we implement a version it can be copy pasted pretty easily most times, and we prevent some annoying footguns.

b-rowan (changes requested):

src/miners/factory/model/mod.rs
s0kil: info does not need to be authenticated

src/miners/backends/vnish/web.rs
s0kil: I'm authenticating with default password

src/miners/factory/mod.rs
```diff
@@ -91,7 +92,9 @@ fn parse_type_from_socket(
         _ if json_string.contains("AVALON") => {
             Some((Some(MinerMake::AvalonMiner), Some(MinerFirmware::Stock)))
         }
-        _ if json_string.contains("VNISH") => Some((None, Some(MinerFirmware::VNish))),
+        _ if json_string.contains("VNISH") => {
+            Some((Some(MinerMake::AntMiner), Some(MinerFirmware::VNish)))
```
b-rowan, line 96: We can't really guarantee this, fine for now but something to think about.

src/miners/factory/mod.rs
```diff
@@ -145,6 +151,9 @@ fn select_backend(
         (Some(MinerMake::BitAxe), Some(MinerFirmware::Stock)) => {
             Some(ESPMiner::new(ip, model?, firmware?, version?))
         }
+        (Some(MinerMake::AntMiner), Some(MinerFirmware::VNish)) => {
```
b-rowan, line 154: In theory this could be `Some(_), Some(MinerFirmware::VNish)`.

src/miners/factory/mod.rs
```diff
@@ -244,14 +253,18 @@ impl MinerFactory {
 
         match miner_info {
             Some((make, firmware)) => {
-                let model = if let Some(miner_make) = make {
+                let model = if let Some(MinerFirmware::VNish) = firmware {
+                    firmware.unwrap().get_model(ip).await
+                } else if let Some(miner_make) = make {
```
b-rowan, line 258: Is this needed?  Should be covered by the lines below, this just adds an extra branch.

src/miners/factory/mod.rs
```diff
@@ -244,14 +253,18 @@ impl MinerFactory {
...
                     miner_make.get_model(ip).await
                 } else if let Some(miner_firmware) = firmware {
                     miner_firmware.get_model(ip).await
                 } else {
                     return Ok(None);
                 };
-                let version = if let Some(miner_make) = make {
+                let version = if let Some(MinerFirmware::VNish) = firmware {
+                    firmware.unwrap().get_version(ip).await
+                } else if let Some(miner_make) = make {
```
b-rowan, line 267: Uneeded?

b-rowan (changes requested):

PR #34, opened 2025-08-07
Update To Latest rust stable

s0kil:


lurkny (approved):
Looks good, the if let chaining will be nice to use.

PR #39, opened 2025-08-11
feature: add BTMiner V1 (M20S)

b-rowan:


lurkny (changes requested):
A couple minor issues...

I know in every other backend we do .get() in loops, but it seems like an easy optimization to avoid repeatedly getting the same data 4-6 times per loop iteration.

src/miners/backends/btminer/v1/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+                let accepted_shares = data
+                    .get(&DataField::Pools)
+                    .and_then(|val| val.pointer(&format!("/{}/Accepted", idx)))
+                    .and_then(|val| val.as_u64());
+
+                let rejected_shares = data
+                    .get(&DataField::Pools)
+                    .and_then(|val| val.pointer(&format!("/{}/Rejected", idx)))
+                    .and_then(|val| val.as_u64());
+
```
lurkny, line 487: Same thing here, maybe move .get() before entering the loop.

lurkny (approved):

PR #57, opened 2025-08-21
feat: Implement Mara FW

### LINKED ISSUE #56: Impl Mara Firmware

lurkny:
Closes #56 

Only things I feel icky on is the firmware detection in the factory, as well as the frequency field in hashboard parsing, we are only given the average frequency, so I was unsure about including it on a per hashboard basis.

### DIFF
--- a/src/data/device/models/antminer.rs
+++ b/src/data/device/models/antminer.rs
@@ -94,7 +94,6 @@ pub enum AntMinerModel {
     #[serde(alias = "s19l")]
     S19L,
     #[serde(alias = "ANTMINER S19 PRO")]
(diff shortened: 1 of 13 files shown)


src/data/device/models/antminer.rs
```diff
@@ -94,6 +94,7 @@ pub enum AntMinerModel {
     #[serde(alias = "s19l")]
     S19L,
     #[serde(alias = "ANTMINER S19 PRO")]
+    #[serde(alias = "ANTMINER S19 PRO (110TH)")]
```
lurkny, line 97: Will fix this soon

b-rowan (approved):

b-rowan (changes requested):

PR #73, opened 2025-09-25
Bitaxe has never been called BitAxe

Georges760:


b-rowan (approved):
PR looks fine, just reached out to Skot to see what his preference is, I just want to confirm with him before merging.

PR #123, opened 2026-01-08
Return best-effort miner identification on timeout (don’t drop Stock matches)

glitchpixelz:
This change prevents miner discovery from returning None when the identification timeout triggers after a valid Stock match was already found (e.g., Bitaxe). We keep the first successful identification as a fallback and, if time runs out, return that best effort result rather than discarding it. Non stock firmware results still take priority and can override Stock when detected. Remaining discovery tasks are aborted once a final decision is made.

Before

- Stock devices could be correctly identified early but still not returned if other probes didn’t complete before the timeout.
- Bitaxe could disappear from scans despite successful AxeOS web detection.

After

- If any valid identification was found before timeout, the best known result is returned.
- Bitaxe is consistently discovered again.
- Non stock firmware remains preferred when detected.

### DIFF
--- a/src/miners/backends/antminer/v2020/mod.rs
+++ b/src/miners/backends/antminer/v2020/mod.rs
@@ -776,11 +776,25 @@ impl Restart for AntMinerV2020 {
 impl Pause for AntMinerV2020 {
     #[allow(unused_variables)]
     async fn pause(&self, at_time: Option<Duration>) -> anyhow::Result<bool> {
-        Ok(self
-            .web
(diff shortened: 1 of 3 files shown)


glitchpixelz:
Not a rebase, the identification logic tried to wait for a better match, a non stock firmware, and could hit the overall identification timeout before all probes finished. When the timeout fired, the function returned None instead of stock. I noticed this thankfully of a bitaxe that I had scanned.

b-rowan:
> Yes that is correct. I committed this to the same branch that was previously merged.

Can you update your master, make a new branch from it, and cherry pick the needed commits?

PR #124, opened 2026-01-08
fix: improve miner discovery

glitchpixelz:
This change prevents miner discovery from returning None when the identification timeout triggers after a valid Stock match was already found (e.g., Bitaxe). We keep the first successful identification as a fallback and, if time runs out, return that best effort result rather than discarding it. Non stock firmware results still take priority and can override Stock when detected. Remaining discovery tasks are aborted once a final decision is made.

Before

- Stock devices could be correctly identified early but still not returned if other probes didn’t complete before the timeout.
- Bitaxe could disappear from scans despite successful AxeOS web detection.

After

- If any valid identification was found before timeout, the best known result is returned.
- Bitaxe is consistently discovered again.
- Non stock firmware remains preferred when detected.

### DIFF
--- a/src/miners/factory/mod.rs
+++ b/src/miners/factory/mod.rs
@@ -64,9 +64,6 @@ async fn check_port_open(ip: IpAddr, port: u16, connectivity_timeout: Duration) 
     // disable Nagle's algorithm for immediate transmission
     let _ = stream.set_nodelay(true);
 
-    // immediate close without waiting for lingering data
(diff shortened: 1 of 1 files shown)


b-rowan (approved):
Tested this against current master and master prior to the priority changes, seems quite consistent.  Functionally seems the same, so no issues there.

Going to approve for now, but I would like to take another look at cleaning this code up slightly in the future, as I find it much more complex to understand.

PR #132, opened 2026-01-29
Fix antminer is_mining to return false when hashrate is zero

glitchpixelz:
Issue: bitmain-work-mode can show normal ("0") even when the miner is overheated and has 0 hashrate, so is_mining ends up true.

Fix: Only report is_mining as true when hashrate is > 0 (when hashrate is available). This keeps the existing work-mode checks and stops the false “mining” state at zero hashrate.

### DIFF
--- a/src/miners/backends/antminer/v2020/mod.rs
+++ b/src/miners/backends/antminer/v2020/mod.rs
@@ -317,7 +317,7 @@ impl GetDataLocations for AntMinerV2020 {
                     tag: None,
                 },
             )],
-            DataField::IsMining => vec![(
+            DataField::MiningMode => vec![(
                 WEB_MINER_CONF,
                 DataExtractor {
                     func: get_by_pointer,
@@ -333,6 +333,7 @@ impl GetDataLocations for AntMinerV2020 {
                     tag: None,
                 },
             )],
+            DataField::IsMining => vec![],
             DataField::Pools => vec![(
                 RPC_POOLS,
                 DataExtractor {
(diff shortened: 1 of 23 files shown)


b-rowan:
I think is_mining is intended to indicate whether the miner SHOULD be mining, not whether or not it truly is.  Might be worth changing the name to mining_enabled or something, but I think representing to the user that this miner is not in sleep mode but just sucks is a different state than having a miner in sleep mode.

Just want to make sure that this aligns with that idea, we can discuss if y'all want to change it.

glitchpixelz:
You're right and thought about this at first and totally undermined the idea with the broad definition of the "stopped" work mode value, but in reality this is "not mining = enabled". Yes the intent is to distinguish if the miner is truly mining or not and it does suck two states as we know of now but also a good thing.

So to be correct we would have to implement something like GetMiningEnabled and would look something like
| MiningEnabled | IsMining | Meaning |
|---|---|---|
| false | false | sleeping / stopped |
| true  | false | should mine but not hashing (overheat, failure, pool issue) |
| true  | true  | normal mining |

b-rowan:
> So to be correct we would have to implement something like GetMiningEnabled

I don't mind this idea, but then isn't `is_mining` just a calculated value directly from hashrate?  Is there another thing we would want to check when calculating this? If not, I think we should leave that check to the end user inside business logic, no reason for us to add a value for `hashrate == 0`...

glitchpixelz:
Added a MiningMode enum (enabled and disabled) to replace original is_mining and kept ismining as "Whether the miner is currently hashing."

src/miners/backends/antminer/v2020/mod.rs
```diff
@@ -526,6 +527,14 @@ impl GetHashrate for AntMinerV2020 {
     }
 }
 
+impl GetIsMining for AntMinerV2020 {
+    fn parse_is_mining(&self, data: &HashMap<DataField, Value>) -> bool {
+        self.parse_hashrate(data)
```
b-rowan: Isn't it better to just compute this in get_data?

glitchpixelz: Yes, this would make it much better.

PR #186, opened 2026-03-18
feat: add sealminer support

b-rowan:


asic-rs-firmwares/sealminer/src/firmware.rs
```diff
@@ -0,0 +1,143 @@
...
+    let client = Client::new();
+
+    let login_response = client
+        .post(format!("http://{ip}/cgi-bin/login.php"))
+        .header(
+            "Content-Type",
+            "application/x-www-form-urlencoded; charset=UTF-8",
+        )
+        .header("X-Requested-With", "XMLHttpRequest")
+        .body("username=seal&origin_pwd=seal")
```
b-rowan: Should use discovery passwords, added since this was opened.

asic-rs-firmwares/sealminer/src/backends/v2025/web.rs
```diff
@@ -0,0 +1,168 @@
...
+use async_trait::async_trait;
+use reqwest::{Client, Method};
+use serde_json::{Value, json};
+
+#[derive(Debug)]
+pub struct SealMinerWebAPI {
+    client: Client,
+    pub ip: IpAddr,
+    auth: MinerAuth,
+}
```
b-rowan, line 19: Could probably store the cookie in a cache here, can likely be reused.

PR #224, opened 2026-04-08
feat(factory): support auth during miner discovery

cfilipescu:
## Summary
- add optional `discovery_auth` to `MinerFactory` so callers can supply credentials during discovery-based miner construction
- pass configured auth to `build_miner` after firmware identification to support stock firmwares requiring non-default digest credentials
- add `with_discovery_auth` builder method and update docs/debug output to reflect configurable discovery auth

### DIFF
--- a/src/factory.rs
+++ b/src/factory.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::HashSet,
+    collections::{HashMap, HashSet},
     net::{IpAddr, Ipv4Addr, SocketAddr},
     panic::AssertUnwindSafe,
     pin::Pin,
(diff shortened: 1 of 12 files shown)


b-rowan (changes requested):

src/factory.rs
```diff
@@ -361,6 +373,18 @@ impl MinerFactory {
         self
     }
 
+    /// Set credentials for a specific firmware entry ID to use during
+    /// miner construction after identification.
+    pub fn with_firmware_discovery_auth(
+        mut self,
+        firmware_id: impl Into<String>,
```
b-rowan: Can't this just be passed in the firmware entry itself?  Firmware should implement display, which implements tostring, so we could remove the firmware ID concept altogether?

b-rowan (approved):

PR #230, opened 2026-04-10
feat: add Auradine support

### LINKED ISSUE #40: Impl Auradine Backend

jpcomps:
## What changed
- add Auradine make and firmware support to the workspace
- implement Auradine v1 RPC and web backends
- register Auradine in miner discovery and factory construction
- add fixture-based tests for Auradine parsing and pool update safety

## Why
This adds first-class support for Auradine miners in `asic-rs`, including discovery, data collection, and control operations.

## Notes
- Auradine pool updates intentionally accept and submit only the first 3 pools
- pool updates require explicit passwords so we do not accidentally blank secrets during round-trip config writes
- mining state now considers the miner sleep state in addition to reported hashrate

## Validation
- `cargo check`
- `cargo test -p asic-rs-firmwares-auradine`
- `cargo test -p asic-rs-makes-auradine`


asic-rs-firmwares/auradine/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1631 @@
...
+}
+
+impl GetPsuFans for AuradineV1 {
+    fn parse_psu_fans(&self, data: &HashMap<DataField, Value>) -> Vec<FanData> {
+        let Some(psu) = data.get(&DataField::PsuFans).and_then(Value::as_object) else {
+            return vec![];
+        };
+
+        let mut fans = Vec::new();
+        for (position, key) in ["FanSpeed1", "FanSpeed2"].into_iter().enumerate() {
```
b-rowan: Should we loop fan count here?  What if the device expects 0 fans?

asic-rs-firmwares/auradine/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1631 @@
...
+                    working_chips: None,
+                    serial_number: None,
+                    chips: vec![],
+                    voltage: None,
+                    frequency: None,
+                    tuned: None,
+                    active: Some(true),
+                });
+            }
+        }
```
b-rowan: Pretty complex logic here, why not just initialize this based off expected boards and expected chips?

asic-rs-firmwares/auradine/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1631 @@
...
+    fn parse_light_flashing(&self, data: &HashMap<DataField, Value>) -> Option<bool> {
+        data.extract_map::<u64, _>(DataField::LightFlashing, |code| code == 3)
+    }
+}
+
+impl GetMessages for AuradineV1 {
+    fn parse_messages(&self, data: &HashMap<DataField, Value>) -> Vec<MinerMessage> {
+        let Some(led) = data.get(&DataField::Messages).and_then(Value::as_object) else {
+            return vec![];
+        };
```
b-rowan, line 1021: So weird to me that this is in a web command called "led", just want to bring attention to this to double check.

asic-rs-firmwares/auradine/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1631 @@
...
+            }
+
+            parsed.push(PoolConfig {
+                url: PoolURL::from(url.to_string()),
+                username: pool
+                    .get("User")
+                    .and_then(Value::as_str)
+                    .unwrap_or_default()
+                    .to_string(),
+                password: String::new(),
```
b-rowan: If password cannot be empty, can this have a default value, like `x` or `123`?  I think we have used `x` in the past.

asic-rs-firmwares/auradine/src/backends/v1/rpc.rs
```diff
@@ -0,0 +1,106 @@
...
+            let message = status_obj.get("Msg").and_then(|v| v.as_str());
+            return Ok(RPCCommandStatus::from_str(status, message));
+        }
+
+        if let Some(status) = value.get("STATUS").and_then(|v| v.as_str()) {
+            return Ok(RPCCommandStatus::from_str(status, None));
+        }
+
+        Ok(RPCCommandStatus::Success)
+    }
```
b-rowan: I think we usually implement a custom trait for this, `trait StatusFromAuradineV1{from_auradine_v1(response:&str)}`

asic-rs-firmwares/auradine/src/backends/v1/web.rs
```diff
@@ -0,0 +1,245 @@
...
+            let message = status_obj.get("Msg").and_then(|v| v.as_str());
+            return Ok(RPCCommandStatus::from_str(status, message));
+        }
+
+        if let Some(status) = value.get("STATUS").and_then(|v| v.as_str()) {
+            return Ok(RPCCommandStatus::from_str(status, None));
+        }
+
+        Ok(RPCCommandStatus::Success)
+    }
```
b-rowan: RPC command status on a web API, I assume because theyre following the CGMiner style with it?  If you implement this parsing on a trait, you should just be able to use that trait.

b-rowan (changes requested):
Looks solid.  So much boilerplate when creating a new type, might be worth trying to make a script to reduce that, but we don't do it that often so I guess its not a big deal.

asic-rs-firmwares/auradine/src/backends/v1/mod.rs
jpcomps: yeah they use it kind of as their status, kind of strange but can display it to the user

jpcomps:
should be close now, some shameful 1 start indexing...so working around that makes it a bit messy but a good start. tested on a machine seems ok

asic-rs-firmwares/auradine/src/backends/v1/status.rs
```diff
@@ -0,0 +1,24 @@
...
+            return Ok(RPCCommandStatus::from_str(status, message));
+        }
+
+        if let Some(status) = self.get("STATUS").and_then(|v| v.as_str()) {
+            return Ok(RPCCommandStatus::from_str(status, None));
+        }
+
+        Ok(RPCCommandStatus::Success)
+    }
+}
```
b-rowan: Usually we put this at the bottom of the RPC file...

Would consider this pattern if we went to something like:
```
rpc/
    mod.rs
    status.rs
```

But maybe a separate PR for that.

b-rowan: Also kindof inverted from what we normally do (see https://github.com/256foundation/asic-rs/blob/965d3dae713e09169fc999945059c3316ade3221/asic-rs-firmwares/antminer/src/backends/v2020/rpc.rs#L139-L160), but I don't mind it.  I think I get why this is like this here, since it's being used in the web API, but maybe something we want to consider migrating to the other types too.

jpcomps:
Addressed the remaining status-parser placement feedback in `6c73479`:

- moved `StatusFromAuradineV1` into `backends/v1/rpc.rs` (matching existing backend pattern),
- reused that parser from the Web API flow,
- removed the separate `backends/v1/status.rs` module.

Also resolved the two open review threads tied to that refactor.

b-rowan (approved):

PR #232, opened 2026-04-12
fix(factory): add bound miner discovery timeouts

jameshilliard:
Hopefully fixes freezing.

### DIFF
--- a/asic-rs-core/src/util.rs
+++ b/asic-rs-core/src/util.rs
@@ -2,7 +2,10 @@ use std::{net::IpAddr, sync::LazyLock, time::Duration};
 
 use reqwest::{StatusCode, header::HeaderMap};
 use serde_json::json;
-use tokio::io::{AsyncRead, AsyncReadExt, AsyncWriteExt};
+use tokio::{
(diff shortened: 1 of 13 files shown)


asic-rs-core/src/util.rs
```diff
@@ -76,15 +92,30 @@ pub async fn read_exact_with_timeout(
...
+        .await
+        .map_err(|_| RPCError::WriteTimeout)?
+        .map_err(RPCError::from)?;
+    Ok(())
+}
+
 #[tracing::instrument(level = "debug")]
 pub async fn send_rpc_command(ip: &IpAddr, command: &'static str) -> Option<serde_json::Value> {
-    let mut stream = tokio::net::TcpStream::connect(format!("{ip}:4028"))
+    let mut stream = connect_tcp_stream((*ip, 4028), DEFAULT_RPC_TIMEOUT)
```
b-rowan, line 110: I think better off passing ip.clone() here.

b-rowan (changes requested):

b-rowan (approved):

PR #251, opened 2026-05-13
fix(antminer-v2020): fix pause/resume on older stock firmware

### LINKED ISSUE #245: AntMinerV2020 Pause::pause does not support older stock set_miner_conf.cgi payload shape
## Summary

  `AntMinerV2020::pause()` currently sends a partial JSON payload to `/cgi-bin/set_miner_conf.cgi`, but at least some older
  Bitmain stock firmware versions require the browser/UI-style full config payload with `Content-Type: text/
  plain;charset=UTF-8`.

  As a result, `pause()` returns `false` or times out, and the miner does not enter sleep mode.

  ## Current behavior

  For stock Antminer firmware, `Pause::pause()` reads `/cgi-bin/get_miner_conf.cgi`, then sends one of these partial JSON
  payloads:

  ```json
  {"miner-mode":"1"}
```
  or:
```json
  {"bitmain-work-mode":"1"}
```
  via AntMinerWebAPI::set_miner_conf.

  On the affected firmware, this does not change the mode. get_miner_conf.cgi continues to report:
```json
  "bitmain-work-mode": "0"
```
  ## Affected device / firmware

  Observed on:

  Antminer S19j Pro
  Linux 4.9.113 #1 SMP PREEMPT Mon Dec 26 17:29:52 CST 2022
  system_filesystem_version: Mon Dec 26 17:10:01 CST 2022
  firmware_type: Release

  ## Browser behavior that works

  The stock web UI sends a full config payload to the same endpoint:

  POST /cgi-bin/set_miner_conf.cgi
  Content-Type: text/plain;charset=UTF-8
  X-Requested-With: XMLHttpRequest
  Accept: application/json, text/javascript, */*; q=0.01

  Example body:
```json
  {
    "bitmain-fan-ctrl": false,
    "bitmain-fan-pwm": "100",
    "miner-mode": 1,
    "freq-level": "100",
    "pools": [
      {
        "url": "stratum+tcp://example:3334",
        "user": "user1",
        "pass": ""
      },
      {
        "url": "stratum+tcp://example:3334",
        "user": "user2",
        "pass": ""
      },
      {
        "url": "stratum+tcp://example:443",
        "user": "user3",
        "pass": ""
      }
    ]
  }
```
  Notable differences from the current asic-rs request:

  - Content-Type is text/plain;charset=UTF-8, not application/json
  - the body is the full miner config, not only the mode key
  - the UI uses miner-mode: 1 as a number
  - the UI includes existing pool/fan/frequency fields

  ## Expected behavior

  AntMinerV2020::pause() should support this older stock UI payload shape, or fall back to it when the partial JSON update
  does not apply.

  A possible approach:

  1. Read get_miner_conf.cgi
  2. Try the current partial JSON behavior
  3. Re-read config and verify sleep mode
  4. If unchanged, send a full UI-style config payload using text/plain;charset=UTF-8
  5. Confirm either miner-mode == 1 or bitmain-work-mode == 1

  ## Why this matters

  Consumers of Pause::pause() expect a successful result to mean the miner is actually entering sleep mode. On this
  firmware, the current call can fail or time out without changing the miner state, preventing safe workflows that need
  mining stopped before installation or maintenance.

taserz:
## Summary

- Older Antminer stock firmware (e.g. S19j Pro Dec 2022) reports the work mode via `bitmain-work-mode` in `get_miner_conf` but `set_miner_conf` only responds to `miner-mode` — and on some builds that write silently no-ops, requiring a further fallback to `bitmain-work-mode`
- The previous implementation always wrote `bitmain-work-mode`, which did nothing on the affected firmware, leaving the miner stuck in its current mode
- Fix: try `miner-mode` first, verify by re-reading `get_miner_conf`, fall back to `bitmain-work-mode` if the write did not apply; skip the whole sequence if the miner is already in the target mode
- Also extends the `set_miner_conf` timeout floor to 35 s (some firmware takes >20 s to respond)

## Test plan

- [x] 10 new unit tests covering all pause/resume code paths via a minimal mock HTTP server
  - Already in target mode (both key variants) → returns `true` without making a POST
  - No mode key present → returns `false`
  - `miner-mode` write succeeds on first try → returns `true`
  - `miner-mode` write accepted (200) but does not apply; fallback to `bitmain-work-mode` succeeds → returns `true`
  - Mirror set for `resume()`
- [x] `cargo test -p asic-rs-firmwares-antminer pause_resume` — all 10 pass

Closes #245

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,6 +53,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ascii-canvas"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8824ecca2e851cec16968d54a01dd372ef8f95b244fb84b84e70128be347c3c6"
+dependencies = [
+ "term",
+]
+
+[[package]]
 name = "asic-rs"
 version = "0.5.1"
 dependencies = [
@@ -131,6 +140,7 @@ dependencies = [
  "chrono",
  "crc32fast",
  "diqwest",
+ "httpmock",
  "macaddr",
  "measurements",
  "once_cell",
@@ -332,7 +342,7 @@ dependencies = [
  "asic-rs-core",
  "asic-rs-makes-whatsminer",
  "async-trait",
- "base64",
+ "base64 0.22.1",
  "chrono",
  "ecb",
  "hex",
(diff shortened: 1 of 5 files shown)


b-rowan:
Superseded by #249

PR #252, opened 2026-05-19
feat: allow selecting data for the `get_data` return

b-rowan:
Also adds the ability to remove `data.hashboards.chips` to account for data constraints, since it can be a very large part of the data and most situations don't require it.


asic-rs-core/src/data/collector.rs
```diff
@@ -1,65 +1,96 @@
...
     FirmwareVersion,
     /// Control board version of the miner.
+    #[cfg_attr(feature = "python", pydantic(value = "ControlBoardVersion"))]
     ControlBoardVersion,
     /// Details about the hashboards (e.g., temperatures, chips, etc.).
+    #[cfg_attr(feature = "python", pydantic(value = "Hashboards"))]
     Hashboards,
+    /// Details about each chip on each hashboard (e.g., temperatures, frequency, etc.).
+    #[cfg_attr(feature = "python", pydantic(value = "Chips"))]
+    Chips,
```
jameshilliard, line 52: Splitting chip payloads into a separate `DataField` means direct `get_hashboards()` calls no longer collect chip data. The default `GetHashboards::get_hashboards()` implementation still calls `collector.collect(&[DataField::Hashboards])`, so backends moved to `DataField::Chips` now return empty `BoardData::chips` via `miner.get_hashboards().await` / Python `Miner.get_hashboards()`, while plain `get_data()` still includes chips. Please update the default to collect both `Hashboards` and `Chips`, or override the affected backends.

asic-rs-firmwares/auradine/src/backends/v1/mod.rs
```diff
@@ -663,6 +690,11 @@ impl GetHashboards for AuradineV1 {
                     .map(str::to_string);
                 hashboard.active = board.get("Enabled").and_then(Value::as_bool);
                 hashboard.tuned = hashboard.active;
+                hashboard.working_chips = match (hashboard.active, hashboard.expected_chips) {
```
jameshilliard: This expected-chip fallback is useful when `DataField::Chips` is excluded, but it also runs when chip data is present. That regresses failed-chip reporting: an active board with dead chips will report all expected chips as working instead of counting parsed chip health like the old code did. Please keep this fallback only for the no-chip-data path and recompute `working_chips` from parsed chips when `chip_data` is present.

asic-rs-firmwares/avalonminer/src/backends/avalon_q/mod.rs
```diff
@@ -486,15 +493,20 @@ impl GetHashboards for AvalonQMiner {
...
 
-                board.intake_temperature = summary["ITemp"][idx]
-                    .as_f64()
-                    .map(Temperature::from_celsius);
+                board.intake_temperature =
+                    summary_f64(summary, "ITemp", idx).map(Temperature::from_celsius);
             }
 
+            board.active = board.hashrate.as_ref().map(|h| h.value > 0.0);
+            board.working_chips = match (board.active, board.expected_chips) {
```
jameshilliard: Same issue here: when `hb_info` is present below, `working_chips` should come from the parsed chip health, not from `active + expected_chips`. Otherwise active boards with failed chips are reported as having every expected chip working. Please use this expected-chip value only as the no-chip-data fallback.

PR #254, opened 2026-05-20
feat: add control functionality needed by proto fleet

b-rowan:
Adds `ReadLogs`, `FactoryReset`, and `ChangePassword` traits for miner firmwares that have meta files.  Added blank implementations for the other ones for now, but this follows the same pattern as the other control traits, adding a function to execute the functions and a function for `supports_{function}`.

Added for epic, braiins, and vnish, since those have API documentation in git.


src/python/miner.rs
```diff
@@ -385,6 +397,33 @@ impl Miner {
...
+        &mut self,
+        py: Python<'a>,
+        password: &str,
+    ) -> PyResult<PyAwaitable<Option<bool>>> {
+        let password = password.to_string();
+        let inner_mut = Arc::get_mut(&mut self.inner).ok_or_else(|| {
+            PyRuntimeError::new_err("cannot update password while miner is in use")
+        })?;
+        let result = pyo3_async_runtimes::tokio::get_runtime()
+            .block_on(inner_mut.change_password(&password));
```
ankitgoswami, line 424: been a while since I worked with async rust, but would a block_on on the current tokio runtime work when awaited?
Also, looks like the Awaitable isn't really needed since the network call runs synchronously before future_into_py.

b-rowan: Good question, I don't have a great way to test this right now, ill have to find a BOS machine or something to test against.

asic-rs-firmwares/braiins/src/backends/v26_04/web.rs
```diff
@@ -0,0 +1,275 @@
...
+            .ok_or(BraiinsError::AuthenticationFailed)
+    }
+
+    pub async fn read_logs(&self) -> anyhow::Result<String> {
+        const LOG_TYPES: &[&str] = &["errors", "bosminer", "boser", "monitor", "syslog", "dmesg"];
+
+        self.ensure_authenticated().await?;
+
+        let mut logs = String::new();
+        for log_type in LOG_TYPES {
```
ankitgoswami: could parallelize these

b-rowan: I think that was supposed to be removed and only load the BOSMiner log.  Whoops...

ankitgoswami (approved):
❤️

src/python/miner.rs
b-rowan: Gonna leave this for now, if we hit issues with it we can deal with them then I guess...

PR #259, opened 2026-05-25
fix: make async cleanup cancellation safe

jameshilliard:
Propagate Python cancellation into Rust futures, keep discovery and RPC work owned by parent futures, and yield during large firmware processing.


b-rowan:
@codex review

b-rowan:
> @codex review

Fine, ill do it myself then...

Damn lazy AI :laughing:

asic-rs-firmwares/epic/src/firmware.rs
```diff
@@ -75,11 +75,18 @@ impl DiscoveryCommands for EPicFirmware {
     }
 }
 
+fn build_discovery_client() -> Result<reqwest::Client, ModelSelectionError> {
+    reqwest::Client::builder()
+        .pool_max_idle_per_host(0)
+        .build()
+        .map_err(|_| ModelSelectionError::NoModelResponse)
+}
```
b-rowan: Is it worth moving all of these to `core::util`?

src/python/miner.rs
```diff
@@ -44,6 +48,52 @@ impl From<Box<dyn MinerTrait>> for Miner {
...
+            supports_read_logs: miner.supports_read_logs(),
+            supports_factory_reset: miner.supports_factory_reset(),
+            supports_pools_config: miner.supports_pools_config(),
+            supports_upgrade_firmware: miner.supports_upgrade_firmware(),
+            supports_scaling_config: miner.supports_scaling_config(),
+            supports_tuning_config: miner.supports_tuning_config(),
+            supports_fan_config: miner.supports_fan_config(),
+        }
+    }
+}
```
b-rowan: What problem is this solving?  This just feels like duplicated data, all the functions and data it references are just directly struct field accesses under the hood, so all this does is duplicate that data somewhere else?

I understand that this is to get around the RwLock so those fields can be accessed when a mutable reference is being held because of async calls from python, but there has to be some way to block waiting for those, since the mutable calls should be nearly instant, such as updating auth information.

b-rowan (approved):

PR #260, opened 2026-05-25
fix(antminer): preserve 2020 miner config updates

DanNicolau:
I found the discrepancy between bitmain-miner-mode and work-mode. Even when the payload is accepted with status 200 OK, the sleep mode would not update. With this change it will now update on old stock firmware. v2023 is left untouched.

Validated on:
- Firmware Version Mon Dec 26 17:10:01 CST 2022


asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
```diff
@@ -77,6 +87,70 @@ fn miner_conf_mode_matches(conf: &Value, mode: MinerMode) -> bool {
...
+    payload.insert(
+        "pools".to_string(),
+        miner_conf
+            .get("pools")
+            .cloned()
+            .unwrap_or_else(|| Value::Array(Vec::new())),
+    );
+
+    payload
+}
```
b-rowan, line 139: I really am not a fan of having to update a whole config at once, something I picked up with pyasic, if you ever get a corrupted config back from the miner (say it hasn't initialized yet), you can just wipe the user's config on the machine, which is not ideal.

Not like this is something I think we can fix, but definitely something to be aware of.

b-rowan (approved):

b-rowan:
@DanNicolau Approved, see comment and lints need fixing though.

asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
DanNicolau: I don't like it either. If I find that they take a partial config somehow I will come back to this and clean it up.

DanNicolau:
changes made

b-rowan:
@DanNicolau not sure what changed, but looks fine.  Couple merge conflicts now from #259, if you can fix I'll merge.

PR #261, opened 2026-05-26
feat: create `MinerComponent` and python bindings

b-rowan:
Adds a `MinerComponent` type for `MinerMessage` to relay what component each message is related to.


asic-rs-firmwares/whatsminer/src/error_codes.rs
```diff
@@ -12,19 +20,111 @@ const UNKNOWN: &str = "Unknown error type.";
...
+fn lookup_component(err_type: u64, err_subtype: u64, err_value: u64) -> Option<MinerComponent> {
+    match err_type {
+        1 => match (err_subtype, err_value) {
+            (1..=3, 0 | 1) => fan_component(err_value),
+            _ => None,
+        },
+        2 => Some(MinerComponent::power_supply(0)),
+        3 => match (err_subtype, err_value) {
+            (0, n) | (2, n) if n != 9 => hashboard_component(n),
+            (2, 9) => Some(MinerComponent::control_board()),
```
ankitgoswami: was the n != 9 guard meant to apply only to (2, n)? I think it will apply to both alternatives making something like 309 fall through to None

asic-rs-firmwares/whatsminer/src/error_codes.rs
```diff
@@ -12,19 +20,111 @@ const UNKNOWN: &str = "Unknown error type.";
 /// * Everything else →
 ///   type = all-but-last-2 digits, subtype = second-to-last digit,
 ///   value = last digit
+#[allow(dead_code)]
 pub fn error_message(code: u64) -> String {
```
ankitgoswami, line 24: only kept for tests? maybe we can use `#[cfg(test)]`

asic-rs-firmwares/whatsminer/src/error_codes.rs
b-rowan: In theory we should never have a hashboard id > 3 (4 boards max), so not sure this is an issue, but I'll look into it.

309 is a PSU code anyway, I'll take a deeper look at the logic here.

ankitgoswami (approved):
thanks!

asic-rs-firmwares/whatsminer/src/error_codes.rs
b-rowan: Much less stupid way to do this, just check for the 329 code first.  Fixed.

asic-rs-firmwares/whatsminer/src/error_codes.rs
b-rowan: Fixed.

PR #266, opened 2026-06-01
feat: add proto backend and rig support

### LINKED ISSUE #265: feat: Implement Proto backend
@ankitgoswami wanted to get your thoughts before it would get implemented, think could be valuable. I do have a rough branch coded against the proto-simulator: https://github.com/jpcomps/asic-rs/tree/codex-proto-rig

jpcomps:
implements #265


asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1271 @@
...
+const WEB_TELEMETRY_FULL: MinerCommand = MinerCommand::WebAPI {
+    command: "/api/v1/telemetry?level=miner,hashboard,psu",
+    parameters: None,
+};
+// Chip-detail endpoints are discovered from hashboard serials, so this
+// internal collector command expands into the real per-ASIC API calls.
+const INTERNAL_CHIP_DETAILS: MinerCommand = MinerCommand::WebAPI {
+    command: "internal:proto/chip-details",
+    parameters: None,
+};
```
b-rowan: Usually put these inside the `GetDataLocations` definition, to keep it all self contained.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1271 @@
...
+const WEB_ERRORS: MinerCommand = MinerCommand::WebAPI {
+    command: "/api/v1/errors",
+    parameters: None,
+};
+const WEB_MINING_TARGET: MinerCommand = MinerCommand::WebAPI {
+    command: "/api/v1/mining/target",
+    parameters: None,
+};
+const WEB_TELEMETRY_FULL: MinerCommand = MinerCommand::WebAPI {
+    command: "/api/v1/telemetry?level=miner,hashboard,psu",
```
b-rowan: Is there a situation where we would prefer this to be considered parameters?  Writable requests still use request body or whatever, but read only GET requests use params for URL params?

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1271 @@
...
+
+        Ok(json!({ "details": details }))
+    }
+}
+
+#[async_trait]
+impl APIClient for ProtoV1 {
+    async fn get_api_result(&self, command: &MinerCommand) -> anyhow::Result<Value> {
+        match command {
+            cmd if cmd == &INTERNAL_CHIP_DETAILS => self.build_chip_payload().await,
```
b-rowan: Not my favorite thing, any better way to do this?  I would like to make `get_api_result` as consistent as possible across types, just a match across command types.

I am also totally fine skipping chip details for now until they add it as a properly documented API endpoint.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
b-rowan: See my comment about `get_api_result`, I think this can be removed, I would prefer well documented endpoints where possible, and I think this is a case where it makes sense to be moved on the proto side rather than us doing a workaround for it.

One of the big benefits of working with a company/team that is open to collaboration!

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1271 @@
...
+                rpm: fan
+                    .get("rpm")
+                    .and_then(Value::as_f64)
+                    .map(AngularVelocity::from_rpm),
+            })
+            .collect()
+    }
+}
+
+impl GetPsuFans for ProtoV1 {}
```
b-rowan, line 609: I feel like this is a thing they should expose.  @ankitgoswami do you know if this is a thing that exists, or another thing we should look at adding?

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1271 @@
...
+}
+
+impl GetTuningTarget for ProtoV1 {
+    fn parse_tuning_target(&self, data: &HashMap<DataField, Value>) -> Option<TuningTarget> {
+        data.extract::<f64>(DataField::TuningTarget)
+            .map(|watts| TuningTarget::Power(Power::from_watts(watts)))
+    }
+}
+
+impl GetLightFlashing for ProtoV1 {}
```
b-rowan, line 628: Another thing I am very surprised to see missing.  @ankitgoswami proto fleet has a light button, is this just missing from the rig?

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1271 @@
...
+}
+
+#[async_trait]
+impl SetPowerLimit for ProtoV1 {
+    async fn set_power_limit(&self, limit: Power) -> anyhow::Result<bool> {
+        self.web
+            .put_json(
+                "/api/v1/mining/target",
+                json!({ "power_target_watts": limit.as_watts().round() as u64 }),
+            )
```
b-rowan: `send_command` + params?  Only issue I can see is with put vs post, but then I think I would prefer to delegate to named functions on the web implementation, such as `web.miner_target()` or something like that.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1271 @@
...
+
+    fn supports_set_power_limit(&self) -> bool {
+        true
+    }
+}
+
+#[async_trait]
+impl Restart for ProtoV1 {
+    async fn restart(&self) -> anyhow::Result<bool> {
+        self.web.post_empty("/api/v1/system/reboot").await?;
```
b-rowan: See above in `SetPowerLimit`, same for these.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1271 @@
...
+
+    fn supports_restart(&self) -> bool {
+        true
+    }
+}
+
+#[async_trait]
+impl Pause for ProtoV1 {
+    async fn pause(&self, _at_time: Option<Duration>) -> anyhow::Result<bool> {
+        self.web.post_empty("/api/v1/mining/stop").await?;
```
b-rowan: See above in SetPowerLimit, same for these.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1271 @@
...
+
+    fn supports_pause(&self) -> bool {
+        true
+    }
+}
+
+#[async_trait]
+impl Resume for ProtoV1 {
+    async fn resume(&self, _at_time: Option<Duration>) -> anyhow::Result<bool> {
+        self.web.post_empty("/api/v1/mining/start").await?;
```
b-rowan: See above in SetPowerLimit, same for these.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1271 @@
...
+                    "username": pool.username,
+                    "password": pool.password,
+                    "priority": idx,
+                })
+            })
+            .collect::<Vec<_>>();
+
+        self.web
+            .put_or_post_json("/api/v1/pools", json!(payload))
+            .await?;
```
b-rowan: See above in `SetPowerLimit`, same for these.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1271 @@
...
+            FanConfig::Auto { target_temp, .. } => json!({
+                "mode": "Auto",
+                "target_temperature_c": target_temp,
+            }),
+            FanConfig::Manual { fan_speed } => json!({
+                "mode": "Manual",
+                "speed_percentage": fan_speed,
+            }),
+        };
+        self.web.put_json("/api/v1/cooling", payload).await?;
```
b-rowan: See above in SetPowerLimit, same for these.

asic-rs-firmwares/proto/src/backends/v1/web.rs
```diff
@@ -0,0 +1,215 @@
...
+            .send_command_with_method(command, Some(payload.clone()), Method::POST)
+            .await
+        {
+            Ok(response) => Ok(response),
+            Err(_) => {
+                self.send_command_with_method(command, Some(payload), Method::PUT)
+                    .await
+            }
+        }
+    }
```
b-rowan: Would like to replace these with named functions for different endpoints.

asic-rs-firmwares/proto/src/test/json/v1/HB-PROTO-SIM-b6bdaf86-0_asics.json
b-rowan, line 1: Internal API, flagging for removal along with the chip parsing until this is a documented API.

asic-rs-firmwares/proto/src/firmware.rs
```diff
@@ -0,0 +1,105 @@
...
+    },
+    util::send_web_command,
+};
+use asic_rs_makes_proto::make::ProtoMake;
+use async_trait::async_trait;
+use serde_json::Value;
+
+// Proto exposes its management API on 8080, so include the port in the
+// discovery path instead of changing shared web discovery for every firmware.
+const WEB_SYSTEM: &str = ":8080/api/v1/system";
```
b-rowan: Any chance there is something we can read from `http://{ip}/` instead?  I really want to avoid adding more identification paths if possible, these are exponential request count increases by network size.

asic-rs-firmwares/proto/src/firmware.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    async fn get_version(ip: IpAddr) -> Option<semver::Version> {
+        let (body, _, _) = send_web_command(&ip, WEB_SYSTEM).await?;
+        let data: Value = serde_json::from_str(&body).ok()?;
+        let version = data
+            .pointer("/system-info/os/version")
+            .and_then(Value::as_str)?;
+        semver::Version::parse(version.trim_start_matches('v')).ok()
+    }
```
b-rowan, line 63: Can make these queries after discovery, need to check against the web root first ideally.

asic-rs-firmwares/proto/src/firmware.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+impl FirmwareIdentification for ProtoFirmware {
+    fn identify_web(&self, response: &WebResponse<'_>) -> bool {
+        response.body.contains("\"product_name\"")
+            && response.body.contains("Proto")
+            && response.body.contains("\"system-info\"")
+    }
```
b-rowan, line 70: May need to change if we can use web root.

asic-rs-makes/proto/src/hardware.rs
```diff
@@ -0,0 +1,58 @@
...
+use crate::models::ProtoModel;
+
+impl From<ProtoModel> for MinerHardware {
+    fn from(model: ProtoModel) -> Self {
+        match model {
+            ProtoModel::Rig => MinerHardware {
+                chips: Some(120),
+                fans: Some(6),
+                boards: Some(9),
+            },
```
b-rowan: I think this has to be done dynamically, imagine a rig-compatible board could have any number of chips, or rig could run with less boards, or a different rig chassis could be developed.  This is really the core part of this that makes it quite complex.

src/factory.rs
```diff
@@ -265,6 +270,9 @@ impl MinerFactory {
             if check_port_open(ip, 80, self.connectivity_timeout).await {
                 return self.get_miner(ip).await;
             }
+            if check_port_open(ip, 8080, self.connectivity_timeout).await {
+                return self.get_miner(ip).await;
+            }
```
b-rowan: Another exponential increase...  Rig should respond on port 80, since there will be a redirect there, so I don't think this is required, expect for the simulator since it doesn't have a web UI.

Another one for @ankitgoswami, any chance of slapping a web UI onto the rig sim?

b-rowan (changes requested):
Proto rig is very difficult because it can support random boards under the same miner, its like trying to treat a container full of different miners like a single miner.  I think we need to be dynamic about constructing hardware information by doing it during initial discovery, but that might be tough.  This means we also need to be have a way for a user to check for hardware changes, since in theory all of this is hot-swappable.

Fun!  😆

asic-rs-firmwares/proto/src/backends/v1/mod.rs
ankitgoswami: the hardware API does provide fan information: https://proto-sdk.github.io/proto-api-docs/#/Hardware/GetHardware

asic-rs-firmwares/proto/src/backends/v1/mod.rs
ankitgoswami: it's here -> https://proto-sdk.github.io/proto-api-docs/#/System/LocateSystem

src/factory.rs
ankitgoswami: this is sorta supported although not packaged with the fake rigs. In the proto-fleet project
```
source ./bin/activate-hermit
cd server/fake-proto-rig && FAKE_RIG_DEFAULT_PASSWORD=Pass123! HTTP_PORT=8080 go run .
cd client && PROXY_URL=http://localhost:8080 npm run dev:protoOS
```

asic-rs-firmwares/proto/src/backends/v1/mod.rs
jpcomps: will prob need to see the rig in practice for this one. didn't see the fans exposed for the PSU, but probably not looking in the right place. we can add this post for sure if its there

jpcomps:
this should be in better shape now @b-rowan @ankitgoswami. let me know what you think. Few things to confirm (port 80 discovery) but the scaffolding should be inline with the other FWs now

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1200 @@
...
+
+// Telemetry is a read-only GET, so the `level` filter is passed as a URL query
+// parameter rather than baked into the command path. It carries parameters, so
+// it can't be a `const` like the other endpoint commands.
+fn web_telemetry_full() -> MinerCommand {
+    MinerCommand::WebAPI {
+        command: "/api/v1/telemetry",
+        parameters: Some(json!({ "level": "miner,hashboard,psu" })),
+    }
+}
```
b-rowan: See https://github.com/256foundation/asic-rs/blob/120e7a7c10086eac9b5850f27bd6f0b3f6ea729c/asic-rs-firmwares/whatsminer/src/backends/v3/mod.rs#L133-L144 for an example of how we have handled this previously.  Can still be inside the trait, but doesn't have to be const.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1200 @@
...
+                self.device_info.hardware.boards = discovered.boards;
+            }
+            if discovered.chips.is_some() {
+                self.device_info.hardware.chips = discovered.chips;
+            }
+            if discovered.fans.is_some() {
+                self.device_info.hardware.fans = discovered.fans;
+            }
+        }
+    }
```
b-rowan: This is fine, I think this can stay as-is for this PR, but in the future I think we will want to introduce some kind of validation for all miner types, something like `trait Validate` which will define a `validate(self) -> bool` value which can be called by an end user to see if some information about that miner has changed and it needs to be rediscovered.  Then the flow will be:
```rust
let miner = MinerFactory().get_miner(IP);
while {
    let data = miner.get_data().await;
    if (!miner.validate().await) {
        // Update miner
        let miner = MinerFactory().get_miner(IP);
    };
}
```

In this case I think we will want to ensure that this type of dynamic hardware information is gathered as part of discovery.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1200 @@
...
+                    web_telemetry_full(),
+                    DataExtractor {
+                        func: get_by_pointer,
+                        key: Some("/hashboards"),
+                        tag: Some("telemetry"),
+                    },
+                ),
+            ],
+            DataField::Hashrate => vec![(
+                web_telemetry_full(),
```
b-rowan: See comment above, this can be moved into this trait, no requirement for it to be const.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1200 @@
...
+    }
+}
+
+impl GetHostname for ProtoV1 {
+    fn parse_hostname(&self, data: &HashMap<DataField, Value>) -> Option<String> {
+        data.get(&DataField::Hostname)
+            .and_then(Value::as_str)
+            .map(ToOwned::to_owned)
+    }
+}
```
b-rowan, line 416: Some of these are a bit verbose compared to what we've done in the past, see:
https://github.com/256foundation/asic-rs/blob/120e7a7c10086eac9b5850f27bd6f0b3f6ea729c/asic-rs-firmwares/whatsminer/src/backends/v3/mod.rs#L351

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1200 @@
...
+        data.extract::<String>(DataField::FirmwareVersion)
+    }
+}
+
+impl GetControlBoardVersion for ProtoV1 {
+    fn parse_control_board_version(
+        &self,
+        data: &HashMap<DataField, Value>,
+    ) -> Option<MinerControlBoard> {
+        data.extract_map::<String, _>(DataField::ControlBoardVersion, MinerControlBoard::known)
```
b-rowan: I think we usually define a control board type for each specific miner, so would like to define a proto specific control boards enum in hardware.rs, then you can do this:
https://github.com/256foundation/asic-rs/blob/120e7a7c10086eac9b5850f27bd6f0b3f6ea729c/asic-rs-firmwares/whatsminer/src/backends/v3/mod.rs#L369-L370

asic-rs-firmwares/proto/src/backends/v1/mod.rs
b-rowan: Will resolve for now, can be a future issue.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1200 @@
...
+                    .and_then(Value::as_str)
+                    .unwrap_or_default();
+                MinerMessage::new(
+                    item.get("timestamp")
+                        .and_then(Value::as_u64)
+                        .unwrap_or_default() as u32,
+                    0,
+                    format!("{source}:{slot} {code_str} {message}"),
+                    MessageSeverity::Warning,
+                )
```
b-rowan: Might need a rebase now that we have a component type that fits in here, should remove this weird format.

Also, why is the error code part of the message?  I assume it has non-int characters in it?  This seems to be a common theme, maybe we should look at making it into a string?

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1200 @@
...
+        }
+    }
+}
+
+#[async_trait]
+impl SetFaultLight for ProtoV1 {
+    async fn set_fault_light(&self, fault: bool) -> anyhow::Result<bool> {
+        // Proto's locate endpoint only flashes the LED for a fixed duration
+        // and auto-expires; there is no way to turn it off early. So enabling
+        // flashes the light, and disabling is a no-op.
```
b-rowan: Interesting.  @ankitgoswami I think this is another one for the rig team to look at, I'll reach out.

asic-rs-firmwares/proto/src/backends/v1/web.rs
```diff
@@ -0,0 +1,297 @@
...
+        &self,
+        command: &str,
+        _privileged: bool,
+        parameters: Option<Value>,
+        method: Method,
+    ) -> Result<Value> {
+        self.send_command_with_method(command, parameters, method)
+            .await
+    }
+}
```
b-rowan, line 278: Doesn't this just make `send_command_with_method` obsolete?  Should be a simple find and replace for it?

asic-rs-firmwares/proto/src/firmware.rs
```diff
@@ -0,0 +1,101 @@
...
+};
+use asic_rs_makes_proto::make::ProtoMake;
+use async_trait::async_trait;
+use serde_json::Value;
+
+// Discovery identifies Proto from its web dashboard on port 80 (the shared
+// `HTTP_WEB_ROOT` path), but the JSON API — including the system endpoint that
+// backs model/version lookups — is served on 8080, so that port is part of the
+// path used after identification.
+const WEB_SYSTEM: &str = ":8080/api/v1/system";
```
b-rowan: Just want to make 100% sure that we know the real rig API runs on :8080?

asic-rs-firmwares/proto/src/firmware.rs
```diff
@@ -0,0 +1,101 @@
...
+    ) -> Result<Box<dyn Miner>, ModelSelectionError> {
+        let model = ProtoFirmware::get_model(ip).await?;
+        let version = ProtoFirmware::get_version(ip).await;
+        let mut miner = crate::backends::ProtoV1::new(ip, model, version);
+        if let Some(auth) = auth {
+            miner.set_auth(auth.clone());
+        }
+        // Proto rigs are heterogeneous; discover the actual hardware layout
+        // during the initial build rather than assuming a fixed shape.
+        miner.refresh_hardware().await;
```
b-rowan: I think we want to move this one level up, and have it here as part of the discovery, then passed into the miner, rather than doing it on the miner struct.  Also might be possible to make the model type take in hardware information so it is contained in the model rather than needing a new param, but probably not a big deal since the `new` function isn't a trait.

asic-rs-makes/proto/src/hardware.rs
```diff
@@ -0,0 +1,59 @@
...
+
+#[derive(Debug, PartialEq, Eq, Clone, Hash, Serialize, Deserialize, Display)]
+pub enum ProtoControlBoard {
+    #[serde(rename = "C1")]
+    C1,
+    #[serde(rename = "C2")]
+    C2,
+    #[serde(rename = "C3")]
+    C3,
+}
```
b-rowan, line 43: Nice, already have the ProtoControlBoard enum defined, make sure to use it in the `GetControlBoard` impl.

b-rowan (commented):
Solid, couple more small things.  Should be very close though other than those nits.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
jpcomps: Yeah I think that makes alot of sense. Could be a few reasons why thats useful. even in situations for example a user changes FW, can revalidate or re-sync

b-rowan: This is actually a known issue on Proto Fleet, part of why I mention this.  Generally I think for long-running monitoring being able to check something like that quickly will be very helpful.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -640,22 +660,28 @@ impl GetMessages for ProtoV1 {
...
-                    format!("{source}:{slot} {code_str} {message}"),
+                    .unwrap_or_default()
+                    .to_string();
+                let timestamp = item
+                    .get("timestamp")
+                    .and_then(Value::as_u64)
+                    .unwrap_or_default() as u32;
+                MinerMessage::with_component(
+                    timestamp,
+                    code,
```
b-rowan, line 678: This is defined as a `u64` , if the result is a string that could be an issue...

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -107,13 +96,18 @@ impl ProtoV1 {
...
                     .max()
-                    .map(|chips| chips as u16)
+                    .or(Some(b.len() as u64))
+                    .map(|count| count as u8)
             }),
+            // TODO: rigs are heterogeneous/sparse, so a single model-level chip
+            // count can't represent them and the core's `chips * boards` would
+            // misreport. Left None for now; real counts are per-`BoardData` and
+            // the working total is in `total_chips`.
+            chips: None,
```
b-rowan: Maybe it makes sense to link chip counts to boards, and just remove the chips structure altogether.  Something to think about now, since you could have `ProtoRigBv1` and `ProtoRigBv2` and they might have different chip counts respectively.  Linking it to a board means we could store static information about each board type, then update the static information for other miners to create them in place, with the rig using pre-defined board types to fill out the hardware info.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -1139,10 +1109,17 @@ mod tests {
...
+            parsed.mac.map(|m| m.to_string()),
+            Some("02:00:00:AD:66:39".to_string())
+        );
         assert_eq!(parsed.serial_number.as_deref(), Some("PROTO-SIM-b6bdaf86"));
         assert_eq!(parsed.expected_hashboards, Some(4));
         assert_eq!(parsed.expected_fans, Some(4));
-        assert_eq!(parsed.expected_chips, Some(480));
+        // Model-level chips is None, so the core's chips*boards aggregate is None;
+        // the real counts live per-board and in total_chips.
+        assert_eq!(parsed.expected_chips, None);
```
b-rowan: I would like this to be populated if possible, see my earlier comment for ideas.

asic-rs-firmwares/proto/src/backends/v1/web.rs
```diff
@@ -28,7 +28,9 @@ impl ProtoWebAPI {
     pub fn new(ip: IpAddr, auth: MinerAuth) -> Self {
         Self {
             ip,
-            port: 8080,
+            // Real rigs serve the API on the standard web port; :8080 is the
+            // simulator's port per the MDK spec.
```
b-rowan: I don't think this comment is needed.

asic-rs-firmwares/proto/src/backends/v1/mod.rs
b-rowan: Can rebase now and use the new boards format, should be much more flexible!

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1367 @@
...
+    }
+
+    fn supports_fan_config(&self) -> bool {
+        true
+    }
+}
+
+impl HasDefaultAuth for ProtoV1 {
+    fn default_auth() -> MinerAuth {
+        MinerAuth::new("", "")
```
b-rowan, line 998: Isn't the default username `root` or `admin` or something?

asic-rs-firmwares/proto/src/backends/v1/mod.rs
```diff
@@ -0,0 +1,1367 @@
...
+    }
+}
+
+impl GetHashrate for ProtoV1 {
+    fn parse_hashrate(&self, data: &HashMap<DataField, Value>) -> Option<HashRate> {
+        data.extract_map::<f64, _>(DataField::Hashrate, |ths| HashRate {
+            value: ths,
+            unit: HashRateUnit::TeraHash,
+            algo: "SHA256".to_string(),
+        })
```
b-rowan: Should `as_unit` this into the default unit, in case that ever changes for some reason, it will be consistent across all miners.

b-rowan (changes requested):

asic-rs-firmwares/proto/src/backends/v1/mod.rs
jpcomps: @ankitgoswami ill confirm with you what this should be

asic-rs-firmwares/proto/src/backends/v1/mod.rs
jpcomps: good catch -- fixed

PR #267, opened 2026-06-04
feat: add scaled tuning target support

cfilipescu:
## Summary
- add scaled tuning target to MinerData and miner backend requirements
- implement default empty scaled tuning target support for existing backends
- parse ePIC PowerPlay scaled tuning target from the lower of Throttle Target and Error Throttle Target

## Tests
- cargo check


jpcomps (approved):
Looks good, I'm assuming for scaled target since ePIC has both Error and Temp scaling possibilities you just are just parsing the lowest one

cfilipescu:
> Looks good, I'm assuming for scaled target since ePIC has both Error and Temp scaling possibilities you just are just parsing the lowest one

Yes. For now just putting both in there.

b-rowan (approved):
LGTM 🚀

PR #268, opened 2026-06-05
feat(epic): parse summary status messages

cfilipescu:
## Summary
- collect ePIC summary data for messages
- parse non-empty Status/Last Error as an error message
- print get_messages output in the ignored live ePIC test
- add regression coverage for summary status Last Error parsing

## Tests
- cargo check

### DIFF
--- a/asic-rs-firmwares/epic/src/backends/v1/mod.rs
+++ b/asic-rs-firmwares/epic/src/backends/v1/mod.rs
@@ -19,10 +19,12 @@ use asic_rs_core::{
         fan::FanData,
         firmware::FirmwareImage,
         hashrate::{HashRate, HashRateUnit},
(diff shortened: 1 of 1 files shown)


b-rowan (approved):
PR is good, but definitely not ideal design for how I would like to read errors.  Would like to look into improving it on the firmware side if possible.  One noticeable issue is that the timestamp is linked to when the data was queried, not when the error was generated, and we don't actually know if that error is still active or not.

cfilipescu:
> PR is good, but definitely not ideal design for how I would like to read errors. Would like to look into improving it on the firmware side if possible. One noticeable issue is that the timestamp is linked to when the data was queried, not when the error was generated, and we don't actually know if that error is still active or not.

agreed this should be improved and we can talk about it next week just to brainstorm ideas.

PR #273, opened 2026-06-10
feat: implement scaled tuning target for braiins

b-rowan:
The current implementation for the newer REST API based backends was using scaled tuning target as the tuning target, which could cause some confusion.

For example, if the power limit was set to 3kw, and scaling had backed it off to 2.4kw, it would show the tuning target as 2.4kw, which was only partially correct.


cfilipescu (approved):

b-rowan:
Fixing tests then ill merge.

PR #274, opened 2026-06-15
feat(antminer): add missing trait implementations for stock AM

b-rowan:
Adds all traits that stock antminer firmware can support, basically everything sans scaling.

Includes some generic fixes for python representations of data types I noticed while testing.


asic-rs-firmwares/antminer/src/backends/v2023_07/mod.rs
```diff
@@ -968,15 +1067,87 @@ impl HasAuth for AntMinerV202307 {
...
+            }
+        };
+
+        let pre = self.web.get_miner_conf().await?;
+        let Some(mode_key) = miner_mode_config_key(&pre) else {
+            anyhow::bail!("No Antminer mining mode field found in miner config")
+        };
+
+        self.web
+            .set_miner_conf(json!({ mode_key: mode.to_string() }))
```
ankitgoswami, line 1112: does this need to be merged with the existing configuration?

asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
```diff
@@ -980,21 +1067,41 @@ impl Resume for AntMinerV2020 {
     }
 }
 
+#[async_trait]
 impl ChangePassword for AntMinerV2020 {
+    async fn change_password(&mut self, password: &str) -> anyhow::Result<bool> {
+        let success = self.web.change_password(password).await?;
+        if success {
+            let username = self.web.username().to_string();
+            self.set_auth(MinerAuth::new(username, password));
```
ankitgoswami: the call could time out after the miner accepted the password change and the would leave us with stale creds. Maybe we should re auth after the change_password call to verify what creds are set.

asic-rs-firmwares/antminer/src/backends/v2023_07/mod.rs
b-rowan: I don't think so for 2023?  I don't have a miner with newer FW so I can't really test, but the pause/resume implementation seems to not care -

https://github.com/b-rowan/asic-rs/blob/add-stock-am-traits/asic-rs-firmwares/antminer/src/backends/v2023_07/mod.rs#L939-L966

ankitgoswami (approved):
❤️ left a few comments but mostly looks good. Can't wait to try it on fleet!

asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
b-rowan: Good idea.  That endpoint is kinda slow (same as everything on stock IG)...

b-rowan:
Ill look at the re-auth idea, its probably not a bad idea to try to deal with those funky edge cases.

ankitgoswami (approved):
thanks!

PR #277, opened 2026-06-17
fix(vnish): recognize spelled-out hydro model + per-board water temps + fluid temperature

pos-ei-don:
## Problem

VNish-flashed hydro miners (e.g. Antminer S19 Pro Hydro, VNish 1.3.3) come back with **zero hashboards** and **no temperatures**.

**Root cause — model detection.** These miners report the model spelled out as `ANTMINER S19 PRO HYDRO`, but the only `S19ProHydro` serde alias is Bitmain's abbreviated `ANTMINER S19 PRO HYD.`. So the model parses to `Unknown`, whose default hardware profile has no board layout → `board_count()` is `None` → `parse_hashboards` builds zero boards (and `expected_hashboards` / chip counts come back `None`). The full `S19ProHydro` profile (`4×180`) already exists in `hardware.rs` — it just never gets selected. (`S19Hydro` already uses the spelled-out `ANTMINER S19 HYDRO`; the Pro/Plus and S21 hydro variants still use the abbreviated `HYD.` form and likely need the same for VNish — only the S19 Pro Hydro is verified here.)

**Secondary (hydro temperatures).** Per-board `intake_temperature`/`outlet_temperature` were both taken from `chip_temp/max`, ignoring the `inlet_water_temp`/`outlet_water_temp` the API reports per chain, and `GetFluidTemperature` was unimplemented.

## Fix

1. Add the spelled-out `ANTMINER S19 PRO HYDRO` alias so the existing `S19ProHydro` hardware profile is selected (boards + chip counts populate).
2. Map `inlet_water_temp` → `intake_temperature`, `outlet_water_temp` → `outlet_temperature`, falling back to `chip_temp/max` for air-cooled models.
3. Implement `GetFluidTemperature` as the warmest per-board `outlet_water_temp` (sourced from `/summary`, since `cooling` is empty in immersion/hydro mode).

## Validation

Verified against a live **Antminer S19 Pro Hydro (VNish 1.3.3)**: `/summary` returns 4 chains with `pcb_temp` / `chip_temp` / `inlet_water_temp` / `outlet_water_temp`. Before the change the model parsed to `Unknown` with 0 boards and `fluid_temperature = None`. I don't have a local Rust toolchain here, so I leaned on the existing patterns — happy to add a unit test or adjust on review.

### DIFF
--- a/asic-rs-firmwares/vnish/src/backends/v1_2_0/mod.rs
+++ b/asic-rs-firmwares/vnish/src/backends/v1_2_0/mod.rs
@@ -253,6 +253,14 @@ impl GetDataLocations for VnishV120 {
                     tag: None,
                 },
             )],
+            DataField::FluidTemperature => vec![(
+                WEB_SUMMARY,
+                DataExtractor {
+                    func: get_by_pointer,
+                    key: Some("/miner/chains"),
+                    tag: None,
+                },
+            )],
             _ => vec![],
         }
     }
@@ -378,11 +386,19 @@ impl GetHashboards for VnishV120 {
                 .pointer("/pcb_temp/max")
                 .and_then(|v| v.as_i64())
                 .map(|t| Temperature::from_celsius(t as f64));
-            board.intake_temperature = chain
-                .pointer("/chip_temp/max")
+            // Hydro miners report per-board water temperatures; air-cooled
+            // models do not, so fall back to the chip temperature there.
+            let chip_max = chain.pointer("/chip_temp/max").and_then(|v| v.as_i64());
+            let intake = chain
+                .pointer("/inlet_water_temp")
                 .and_then(|v| v.as_i64())
-                .map(|t| Temperature::from_celsius(t as f64));
-            board.outlet_temperature = board.intake_temperature;
+                .or(chip_max);
+            let outlet = chain
+                .pointer("/outlet_water_temp")
+                .and_then(|v| v.as_i64())
+                .or(intake);
+            board.intake_temperature = intake.map(|t| Temperature::from_celsius(t as f64));
+            board.outlet_temperature = outlet.map(|t| Temperature::from_celsius(t as f64));
             board.working_chips = chain
                 .pointer("/chips")
                 .and_then(|v| v.as_array())
@@ -525,7 +541,21 @@ impl GetFans for VnishV120 {
 
 impl GetPsuFans for VnishV120 {}
 
-impl GetFluidTemperature for VnishV120 {}
+impl GetFluidTemperature for VnishV120 {
+    fn parse_fluid_temperature(&self, data: &HashMap<DataField, Value>) -> Option<Temperature> {
+        // Fluid temperature mirrors other firmwares' "environment temperature":
+        // the coolant entering the machine. For hydro miners that's the
+        // per-board inlet water temperature.
+        let chains = data
+            .get(&DataField::FluidTemperature)
+            .and_then(|v| v.as_array())?;
+        chains
+            .iter()
+            .filter_map(|c| c.pointer("/inlet_water_temp").and_then(|v| v.as_i64()))
+            .max()
+            .map(|t| Temperature::from_celsius(t as f64))
+    }
+}
 
 impl GetWattage for VnishV120 {
     fn parse_wattage(&self, data: &HashMap<DataField, Value>) -> Option<Power> {
--- a/asic-rs-makes/antminer/src/models.rs
+++ b/asic-rs-makes/antminer/src/models.rs
@@ -89,8 +89,10 @@ pub enum AntMinerModel {
     #[serde(alias = "ANTMINER S19 HYDRO")]
     S19Hydro,
     #[serde(alias = "ANTMINER S19 PRO HYD.")]
+    #[serde(alias = "ANTMINER S19 PRO HYDRO")]
     S19ProHydro,
     #[serde(alias = "ANTMINER S19 PRO+ HYD.")]
+    #[serde(alias = "ANTMINER S19 PRO+ HYDRO")]
     S19ProPlusHydro,
     #[serde(alias = "ANTMINER S19K PRO")]
     S19KPro,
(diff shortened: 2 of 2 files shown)


asic-rs-firmwares/vnish/src/backends/v1_2_0/mod.rs
```diff
@@ -324,12 +332,21 @@ impl GetHashboards for VnishV120 {
...
+            Some(n) if n > 0 => n as usize,
+            _ => all_chains
+                .iter()
+                .filter_map(|c| c.pointer("/id").and_then(|v| v.as_u64()))
+                .max()
+                .unwrap_or(0) as usize,
+        };
+        let mut hashboards: Vec<BoardData> = (0..board_count)
+            .map(|idx| BoardData::new(idx as u8, self.device_info.hardware.chips_for_board(idx)))
+            .collect();
```
b-rowan: My preference here is to add the relevant miner type information rather than parse based on API responses.  This solution is probably fine to keep for future implementations, but missing information is usually the symptom of a different problem, and it encourages users to tell our team about it so we can fix it for everyone.

asic-rs-firmwares/vnish/src/backends/v1_2_0/mod.rs
```diff
@@ -525,7 +550,20 @@ impl GetFans for VnishV120 {
...
+impl GetFluidTemperature for VnishV120 {
+    fn parse_fluid_temperature(&self, data: &HashMap<DataField, Value>) -> Option<Temperature> {
+        // Hydro miners circulate coolant; expose the warmest per-board outlet
+        // water temperature as the miner-level fluid temperature.
+        let chains = data
+            .get(&DataField::FluidTemperature)
+            .and_then(|v| v.as_array())?;
+        chains
+            .iter()
+            .filter_map(|c| c.pointer("/outlet_water_temp").and_then(|v| v.as_i64()))
```
b-rowan: I think we want to use inlet here, this is a bit of a tricky one, but this is what other firmwares would call "environment temperature".  EG whatsminers this is measured by a sensor on their control board.  The goal is to illustrate the temperature of the air or water coming into the machine, to be able to see how it changes when it reaches the boards and chips.
```suggestion
            .filter_map(|c| c.pointer("/inlet_water_temp").and_then(|v| v.as_i64()))
```

b-rowan (changes requested):
A few small things, but I am a bit concerned since the S19ProHydro is already defined and has hardware information.  I would like to find out WHY it isn't picking that up in your case, but that could possibly be opened as a separate issue if needed.

See hardware definition here - https://github.com/256foundation/asic-rs/blob/43d706ef31aa154f8bea0bd7ef8e81caa9cbe491/asic-rs-makes/antminer/src/hardware.rs#L176-L179

asic-rs-makes/antminer/src/models.rs
```diff
@@ -89,6 +89,7 @@ pub enum AntMinerModel {
     #[serde(alias = "ANTMINER S19 HYDRO")]
     S19Hydro,
     #[serde(alias = "ANTMINER S19 PRO HYD.")]
+    #[serde(alias = "ANTMINER S19 PRO HYDRO")]
```
b-rowan, line 92: Figures.  Can you apply this same fix to the other hydro miners?  Good likelihood the other types have the same issue.

pos-ei-don:
Validated end-to-end against a live **Antminer S19 Pro Hydro (VNish 1.3.3)**: built a wheel from this branch and ran `MinerFactory().get_miner()` + `get_data()` against the miner.

- **Before:** 0 hashboards, `expected_hashboards = None`, `fluid_temperature = None` (model parsed to `Unknown`).
- **After:** model resolves to `S19ProHydro` (`expected_hashboards = 4`), **4 hashboards** populated with per-board water temps — `intake_temperature = inlet_water_temp`, `outlet_temperature = outlet_water_temp` (≈31–32 °C in / 44–46 °C out) — and `fluid_temperature = 46 °C` (warmest outlet). Per-board hashrates ≈35–37 TH/s.

All three changes confirmed working on real hardware.

b-rowan:
Perfect.  Can you fix the other issues from the comments and then should be good to merge.

pos-ei-don:
Thanks for the quick review! Addressed all of it:

1. **Board count from hardware vs API parsing** — agreed; the chain-id fallback is already removed in the current revision. The real fix is the model alias (below), so the existing `S19ProHydro` hardware profile (4×180) gets selected — no API-derived board count anymore.

2. **Why the profile wasn't picked up** — it's purely the model string. VNish reports `ANTMINER S19 PRO HYDRO` (spelled out), but the only `S19ProHydro` alias was Bitmain's abbreviated `ANTMINER S19 PRO HYD.`, so it parsed to `Unknown` → default profile → 0 boards. The profile itself was correct, it just wasn't being matched. Adding the spelled-out alias selects it.

3. **Other hydro types** — done. Added the spelled-out `… HYDRO` alias to S19 Pro+, S21, S21+ and S21E XP hydros as well (S19 Hydro already had it). Heads-up: only the S19 Pro Hydro is verified on real hardware here; the others follow the same `HYD.` → `HYDRO` pattern, so worth a sanity check if anyone has those on VNish.

4. **fluid_temperature** — switched to `inlet_water_temp` per your note (incoming / environment temperature).

I don't have a local Rust toolchain on this box, so I'm relying on CI for the build — happy to tweak anything else.

b-rowan (approved):
LGTM 🚀

pos-ei-don:
Thanks for the fast review and merge! 🙏

I'll rebuild against `0.6.1` once it's tagged and re-confirm on the S19 Pro Hydro to make sure nothing regressed. The other hydro aliases (S19 Pro+/S21/S21+/S21E XP) are still unverified on real hardware on my side — if anyone shows up with one of those on VNish, happy to help sanity-check the `… HYDRO` aliases against an actual device.

Separately: VNish power/preset writes are still a stub in the VNish backend (`set_power_limit`). I've got a working REST-based path downstream and would be glad to open a PR to wire native VNish power-limit/preset writes into asic-rs if that's something you'd want — just say the word and I'll file an issue to scope it.

PR #282, opened 2026-06-17
fix(pydantic): serialize Duration as timedelta, not float

pos-ei-don:
### Problem
`MinerData` fields typed `Option<Duration>` (notably `uptime`) are serialized to Python as a **bare float (seconds)**: `to_pydantic_data` calls `duration_to_seconds()` and the serialization schema is `float_schema`. Consumers that treat the value as a duration crash:

```
'float' object has no attribute 'total_seconds'
```

This is also **inconsistent within the library**:
- `from_pydantic` already accepts a `timedelta` (it's the first form in its own error message: *"Expected duration as timedelta, …"*).
- The `Miner.get_uptime()` method already returns `Optional[datetime.timedelta]` (its docstring says so).

So only the `MinerData.uptime` **field** comes through as a float, while the equivalent method returns a `timedelta`.

### Fix
Emit a `datetime.timedelta` from `to_pydantic_data` (pyo3 converts `std::time::Duration` natively) and use `timedelta_schema` for serialization. `from_pydantic` is unchanged and still accepts `timedelta`, non-negative seconds, or `{secs}` dict — so existing inputs keep working; only the output type becomes the (already-expected) `timedelta`.

### Verified
Built a wheel with this change and ran it under Home Assistant against a live miner (BraiinsOS S19k Pro). Before: the `uptime` sensor crashed every update with the `total_seconds` error. After: `uptime` populates normally (e.g. `1308 s`) and no errors in the log; other fields unaffected.

### Note
This changes the serialized type of all `Duration` fields from `float` to `timedelta`. That's the type `from_pydantic`/`get_uptime` already imply, but flagging it in case any consumer relied on the float form.

### DIFF
--- a/asic-rs-pydantic/src/lib.rs
+++ b/asic-rs-pydantic/src/lib.rs
@@ -183,10 +183,6 @@ impl PyPydanticType for MacAddr {
     }
 }
 
-fn duration_to_seconds(duration: Duration) -> f64 {
-    duration.as_secs() as f64
-}
-
 impl PyPydanticType for Duration {
     fn pydantic_schema<'py>(
         core_schema: &Bound<'py, PyAny>,
@@ -194,7 +190,7 @@ impl PyPydanticType for Duration {
     ) -> PyResult<Bound<'py, PyAny>> {
         match mode {
             PydanticSchemaMode::Validation => core_schema.call_method0("any_schema"),
-            PydanticSchemaMode::Serialization => core_schema.call_method0("float_schema"),
+            PydanticSchemaMode::Serialization => core_schema.call_method0("timedelta_schema"),
         }
     }
 
@@ -218,10 +214,11 @@ impl PyPydanticType for Duration {
     }
 
     fn to_pydantic_data(&self, py: Python<'_>) -> PyResult<Py<PyAny>> {
-        Ok(duration_to_seconds(*self)
-            .into_pyobject(py)?
-            .into_any()
-            .unbind())
+        // Expose Duration as a Python `datetime.timedelta` (pyo3 native), so it
+        // round-trips with `from_pydantic` (which already accepts timedelta) and
+        // matches the `get_uptime()` method. Previously serialized as a bare
+        // float (seconds), which broke `.total_seconds()` on consumers.
+        Ok((*self).into_pyobject(py)?.into_any().unbind())
     }
 }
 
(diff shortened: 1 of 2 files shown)


asic-rs-pydantic/src/lib.rs
```diff
@@ -218,10 +214,11 @@ impl PyPydanticType for Duration {
...
-            .into_pyobject(py)?
-            .into_any()
-            .unbind())
+        // Expose Duration as a Python `datetime.timedelta` (pyo3 native), so it
+        // round-trips with `from_pydantic` (which already accepts timedelta) and
+        // matches the `get_uptime()` method. Previously serialized as a bare
+        // float (seconds), which broke `.total_seconds()` on consumers.
+        Ok((*self).into_pyobject(py)?.into_any().unbind())
     }
 }
```
b-rowan, line 223: Might be able to remove this whole impl, not sure.  Something to test.  If not, then this is fine as-is.

pos-ei-don:
Quick question on direction here, since there's an internal inconsistency I'd rather you decide on than assume:

- The `uptime` **type stub** declares `timedelta | None`, and the **`get_uptime()` method** returns a `timedelta` (its docstring even says so).
- But the **pydantic field** serialized a `Duration` as a bare **float (seconds)** — and there was a test (`test_miner_data_serializes_uptime_seconds`) asserting exactly that.

So the declared/typed contract (timedelta) and the runtime serialization (float) disagree. This PR resolves it toward **timedelta** (and I updated the test accordingly), which also matches how the sibling rich types serialize as structured objects rather than bare scalars.

But it's your call — two clean options:
1. **timedelta** (this PR): field matches the stub + `get_uptime()`; consumers can use `.total_seconds()`.
2. **float seconds** (your existing test): keep the serialization, and I'll drop this PR and adjust the downstream consumer to read the float instead.

Happy to go either way — just didn't want to silently flip a behavior you'd written a test for. Which do you prefer?

b-rowan:
I think this is the right direction, time delta seems like the right way to represent it.  There may be a situation where we want to implement `model_dump_json` where this might get more difficult (because we would need to dump duration to a float there, then ensure we can still load it), but for now this makes sense.

pos-ei-don:
Thanks — agreed, `timedelta` keeps it symmetric with the stub and `get_uptime()`. Re your `model_dump_json` note: that's a separate, future concern (we'd convert the Duration to seconds on dump and parse it back on load) and doesn't block this change. I kept the small `impl` since removing it drops back to the bare-float serialization. Ready to merge whenever you're happy with it.

b-rowan (approved):
tACK, tested pydantic model dump vs a custom model with timedelta, same functionality.

PR #284, opened 2026-06-18
feat(vnish): implement SetPowerLimit (preset-based)

### LINKED ISSUE #280: feat: implement `SetPowerLimit` for vnish
As mentioned in https://github.com/256foundation/asic-rs/pull/277#issuecomment-4734134526

pos-ei-don:
Closes #280.

VNish has no arbitrary-wattage power control — it exposes named, tuned autotune presets (each ~ a wattage) plus a throttle. So `set_power_limit(limit)` maps the requested watts to the **highest tuned preset whose power is `<= limit`** (same approach pyasic uses), sets `overclock.preset` to it and verifies the change. `supports_set_power_limit()` returns `true`.

- Adds two web helpers: `settings()` (GET) and `autotune_presets()` (GET).
- Preset power is parsed from the preset's `pretty` field (`"3495 watt ~ 132 TH"`); only `status == "tuned"` presets are considered. If no tuned preset fits under the limit, returns `Ok(false)` (no change).
- The existing `web.set_settings()` + verify path is reused.

**Throttle** (percent-of-full-power) is intentionally left out of `set_power_limit` — it's a separate axis; happy to propose it separately as a VNish-specific capability if there's interest.

### Tested
Live on an Antminer S19 Pro Hydro (VNish 1.3.3) via Home Assistant:
- `set_power_limit(3700 W)` → preset `3635` (highest tuned ≤ 3700)
- `set_power_limit(3500 W)` → preset `3495`

both applied and verified against the miner. (I don't have a local Rust toolchain, so I'm relying on CI for the build — happy to adjust anything.)

Context: we needed this for our own deployment but aren't asic-rs architects — please review/change/reject the approach as you see fit.

### DIFF
--- a/asic-rs-firmwares/vnish/src/backends/v1_2_0/mod.rs
+++ b/asic-rs-firmwares/vnish/src/backends/v1_2_0/mod.rs
@@ -710,7 +710,60 @@ impl SetFaultLight for VnishV120 {
 #[async_trait]
 impl SetPowerLimit for VnishV120 {
     fn supports_set_power_limit(&self) -> bool {
-        false
+        true
+    }
+
+    /// VNish sets power by selecting a tuned autotune preset, not an arbitrary
+    /// wattage (mirrors pyasic's behaviour). Pick the highest tuned preset whose
+    /// power is <= the requested limit, set `overclock.preset` to it and verify.
+    async fn set_power_limit(&self, limit: Power) -> anyhow::Result<bool> {
+        let target_watts = limit.as_watts() as i64;
+
+        let presets = self.web.autotune_presets().await?;
+        let preset_list = presets
+            .as_array()
+            .or_else(|| presets.get("presets").and_then(|p| p.as_array()))
+            .ok_or_else(|| anyhow::anyhow!("unexpected /autotune/presets response"))?;
+
+        // Each preset: { "name": "3495", "pretty": "3495 watt ~ 132 TH",
+        //               "status": "tuned", ... }. Power is parsed from `pretty`.
+        let best = preset_list
+            .iter()
+            .filter(|p| p.get("status").and_then(|s| s.as_str()) == Some("tuned"))
+            .filter_map(|p| {
+                let pretty = p.get("pretty")?.as_str()?;
+                let (watt_part, _) = pretty.split_once('~')?;
+                watt_part.replace("watt", "").trim().parse::<i64>().ok()
+            })
+            .filter(|&w| w <= target_watts)
+            .max();
+
+        let Some(preset_watts) = best else {
+            return Ok(false);
+        };
+
+        let mut settings = self.web.settings().await?;
+        {
+            let overclock = settings
+                .pointer_mut("/miner/overclock")
+                .ok_or_else(|| anyhow::anyhow!("settings missing miner.overclock"))?;
+            overclock["preset"] = json!(preset_watts.to_string());
+        }
+        let overclock = settings
+            .pointer("/miner/overclock")
+            .cloned()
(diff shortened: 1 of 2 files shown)


asic-rs-firmwares/vnish/src/backends/v1_2_0/mod.rs
```diff
@@ -710,7 +710,61 @@ impl SetFaultLight for VnishV120 {
...
+        let preset_list = presets
+            .as_array()
+            .or_else(|| presets.get("presets").and_then(|p| p.as_array()))
+            .ok_or_else(|| anyhow::anyhow!("unexpected /autotune/presets response"))?;
+
+        // Each preset: { "name": "3495", "pretty": "3495 watt ~ 132 TH",
+        //               "status": "tuned", ... }. Power is parsed from `pretty`.
+        let best = preset_list
+            .iter()
+            .filter(|p| p.get("status").and_then(|s| s.as_str()) == Some("tuned"))
```
b-rowan: Is there a situation where we want to use an un-tuned preset?  What is the difference between them?

I assume tuned presets can be switched to instantly, whereas untuned ones need to run the tuning cycle, and so may require a full boot?  In that case, I think allowing untuned presets as well makes sense, this is the same behavior as other types have (EG Whatsminer always reboots when setting a power limit).

b-rowan (commented):
Tentatively approved, but want to confirm the functionality of VNish (see my comment).

pos-ei-don:
Good point. On VNish: **tuned** presets switch over more or less instantly; an **un-tuned** preset triggers a tuning cycle, i.e. effectively a restart before it settles. Right now the PR only picks the highest *tuned* preset ≤ target watts. I'll broaden it to consider **all** presets (tuned + un-tuned), picking the best ≤ watts and accepting the restart for un-tuned ones — same idea as Whatsminer always rebooting on a power-limit change.

On the integration side I'll also surface the miner's tuning state in its status text, so it's clear why the hashrate ramps after an un-tuned preset rather than looking like a fault. The miner's powered down at the moment, so I'll verify the un-tuned→restart behaviour on the S19 Pro Hydro and push the update shortly.

b-rowan:
> On the integration side I'll also surface the miner's tuning state in its status text, so it's clear why the hashrate ramps after an un-tuned preset rather than looking like a fault. The miner's powered down at the moment, so I'll verify the un-tuned→restart behaviour on the S19 Pro Hydro and push the update shortly.

Not 100% sure how to track this, I think we can use the `tuned` flag on the hashboards data structure to represent it.  Generally I think for most users this is expected behavior, but will get better in repeated setups using the same duplicated power limit over and over.  Fully "dynamic", rebootless tuning is a relatively new concept, so most people likely expect a reboot anyway, but I understand given the specifics of your setup why you would want to avoid reboots.

pos-ei-don:
Pushed the update so `set_power_limit` considers all presets, not just the tuned ones, and reads the wattage from the preset `name` instead of parsing `pretty`.

Verified it against my live S19 Pro Hydro (VNish 1.3.4) — the full `/autotune/presets` list:

- Every real preset has `name` = the bare watt integer, for both **tuned** (3495–5560) and **un-tuned** (5725–7700). Only `"disabled"` is non-numeric, and the `name.parse::<i64>().ok()` in the `filter_map` skips it.
- With the old `status == "tuned"` filter a target of e.g. 6000 W would silently cap at 5560 (the highest tuned preset); now it correctly selects 6000, which just kicks off a tuning cycle before it settles.

CI is recompiling now.

b-rowan (approved):

PR #295, opened 2026-06-23
feat(timezone): get / set / list miner timezone

pos-ei-don:
**Why I need this:** on **VNish** the timezone is a fixed UTC offset with no DST handling, so it has to be switched **by hand twice a year** (summer/winter) — and that gets forgotten, leaving the miner's clock (and its logs/timestamps) an hour off for months. Keeping a device's clock correct across DST is exactly the kind of thing a controller like Home Assistant should own, so I want to be able to read and set the timezone from there instead of remembering to do it on each miner.

This adds a `SetTimezone` capability (get / set / list + supports) so a consumer can do that:

- New `SetTimezone` trait (default: unsupported), added to the `Miner` supertrait + blanket impl; all backends implement it (trivial except the two below).
- **BraiinsOS** 26.04: named zones via `bos.timezone` / `timezoneList` / `setTimezone` — once a named zone is set, BOS handles DST itself.
- **VNish**: a fixed UTC offset via `/settings` `regional.timezone` (read-modify-write); since VNish has no DST awareness, the consumer (re)applies the correct offset at each DST change.
- Python: `get_timezone` / `list_timezones` / `set_timezone` + `supports_set_timezone`.

CI-green; live write-test pending on my fleet. Happy to shape the API (e.g. a typed timezone, or naming) to your preference.


cfilipescu (approved):

asic-rs-core/src/traits/miner.rs
```diff
@@ -41,12 +41,24 @@ pub trait MinerConstructor {
 }
 
 pub trait Miner:
-    GetMinerData + HasMinerControl + SupportsConfigs + UpgradeFirmware + HasAuth + HasDefaultAuth
+    GetMinerData
+    + HasMinerControl
+    + SupportsConfigs
+    + UpgradeFirmware
+    + SetTimezone
```
b-rowan: Can this be moved under `SupportsConfigs`?  Then we can have a `get_timezone_config`/`set_timezone_config`.

asic-rs-core/src/traits/miner.rs
```diff
@@ -745,6 +757,28 @@ pub trait SetPowerLimit {
...
+        anyhow::bail!("Setting timezone is not supported on this platform");
+    }
+    /// The miner's currently configured timezone, if known.
+    async fn get_timezone(&self) -> anyhow::Result<Option<String>> {
+        Ok(None)
+    }
+    /// The timezones the miner accepts (named zones where available; may be empty).
+    async fn list_timezones(&self) -> anyhow::Result<Vec<String>> {
+        Ok(Vec::new())
+    }
```
b-rowan: Does it make sense to use some kind of built in from chrono or something, rather than a list from the miner, which can then be converted to a miner-side tz?  We know there should be a certain number of time zones...

b-rowan (changes requested):

asic-rs-core/src/traits/miner.rs
b-rowan: Maybe not, just use a default list of timezones as a fallback.

pos-ei-don:
Re the list (your follow-up): agreed — I'll use the miner's own list where it has one and fall back to a built-in default list otherwise, no `chrono` needed.

Re moving it under `SupportsConfigs`: makes sense, I'll refactor into the config model as `get_timezone_config`/`set_timezone_config` to match the other configs (same shape as the temperature config). I'll push that rework.

For context on the end goal: this is so Home Assistant can keep a miner's timezone correct across the DST changeover automatically — VNish has no DST handling, so by hand it just gets forgotten.

asic-rs-core/src/config/timezone.rs
```diff
@@ -0,0 +1,55 @@
...
+                    .transpose()?
+                    .flatten(),
+                available: get_optional_field(&obj, "available")?
+                    .map(|value| value.extract())
+                    .transpose()?
+                    .unwrap_or_default(),
+            })
+        }
+    }
+}
```
b-rowan, line 55: Should be able to remove this, automatic with the pydantic macros, reference here:
https://github.com/256foundation/asic-rs/blob/38552b181dbd28e48826da957754fb1c4d1a29e9/asic-rs-core/src/config/pools.rs#L7-L24

asic-rs-core/src/config/timezone.rs
```diff
@@ -0,0 +1,55 @@
+#[cfg(feature = "python")]
+use pyo3::prelude::*;
+use serde::{Deserialize, Serialize};
+
+#[cfg_attr(
+    feature = "python",
+    pyclass(skip_from_py_object, get_all, module = "asic_rs")
```
b-rowan, line 7: Drop `skip_from_py_object`

asic-rs-firmwares/braiins/src/backends/v26_04/mod.rs
```diff
@@ -606,6 +620,58 @@ impl GetScaledTuningTarget for BraiinsV2604 {
...
+            .graphql
+            .send_graphql_command(mutation, true, Some(variables))
+            .await?;
+        // BosResult is a union; a `BosError` variant signals failure.
+        let typename = result
+            .pointer("/bos/setTimezone/__typename")
+            .and_then(|v| v.as_str());
+        Ok(matches!(typename, Some(t) if t != "BosError"))
+    }
+}
```
b-rowan, line 702: Copy paste this impl for every braiins version , for some reason this isn't in the REST api.

asic-rs-firmwares/epic/src/backends/v1/mod.rs
```diff
@@ -169,6 +169,7 @@ impl GetConfigsLocations for PowerPlayV1 {
                     tag: None,
                 },
             )],
+            ConfigField::Timezone => vec![],
```
b-rowan: Doesn't need to exist?  Rather a `_ => vec![],`
```suggestion
            _ => vec![],
```

asic-rs-firmwares/vnish/src/backends/v1_2_0/mod.rs
```diff
@@ -551,6 +566,52 @@ impl GetFans for VnishV120 {
...
+    /// and it has no list endpoint, so we fall back to the whole-hour offsets.
+    fn parse_timezone_config(
+        &self,
+        data: &HashMap<ConfigField, Value>,
+    ) -> anyhow::Result<TimezoneConfig> {
+        const DEFAULT_OFFSETS: &[&str] = &[
+            "GMT-12", "GMT-11", "GMT-10", "GMT-9", "GMT-8", "GMT-7", "GMT-6", "GMT-5", "GMT-4",
+            "GMT-3", "GMT-2", "GMT-1", "GMT+0", "GMT+1", "GMT+2", "GMT+3", "GMT+4", "GMT+5",
+            "GMT+6", "GMT+7", "GMT+8", "GMT+9", "GMT+10", "GMT+11", "GMT+12", "GMT+13", "GMT+14",
+        ];
```
b-rowan, line 597: Do we pull in any "real" timezone objects in any the deps?  Would be nice to be able to go to/from a properly structure timezone object for this, get more type hinting rather than just strings.

b-rowan (changes requested):

pos-ei-don:
Rebased onto v0.8.0 — only the generated .pyi needed a merge; kept the upstream set_tuning_config signature; added an empty SupportsTimezoneConfig impl for the new elphapex backend. cargo check is green.

b-rowan:
Couple comments to address from my old review.

pos-ei-don:
Rebased onto current master. The conflict came from #291 landing earlier today: both PRs add a trait to the `Miner` supertrait and an empty impl to the same backends, so every collision was mechanical — both sides kept, nothing dropped.

**`skip_from_py_object` / the hand-written converter** — both done in `24c4806`. `TimezoneConfig` now has the same shape as `config/scaling.rs`: `from_py_object` + `py_pydantic_model`, manual `py_new` kept, no manual `FromPyObject`.

**Real timezone objects instead of strings** — I'd like to push back on this one, because the string isn't laziness. VNish doesn't have named zones at all: it stores a fixed UTC offset under `/regional/timezone` as `"GMT+2"`, doesn't track DST, and has no list endpoint (hence the hardcoded fallback list). BraiinsOS does give you real IANA zones via `timezoneList`. So `chrono-tz` would type the Braiins side properly and leave VNish with nowhere to go — the `String` is the common denominator of two firmwares, not a shortcut.

If you want the type safety anyway, the honest shape is probably an enum rather than a single string — something like `Named(String)` for IANA zones and `Offset(i8)` for VNish-style fixed offsets, so the two can't be confused and each backend declares which it speaks. Happy to build that if you like it; it changes the public type, so I'd rather have your call than guess.

**Copying the impl to every Braiins version** — I'd like to check something first. Right now only `v26_04` has the real implementation; `v21_09`/`v25_03`/`v25_05`/`v25_07` carry empty impls. Copying it is easy, but each also needs a `ConfigField::Timezone` location and a `send_graphql_command` call, which today only `v26_04` uses. Do the older BOS releases actually serve `bos { setTimezone }` and `timezone`/`timezoneList`? I can only test against BraiinsOS on one machine here, so if I copy it blind I'd be declaring `supports_timezone_config() == true` on four backends I can't verify — which is the sort of thing that reads as supported and then silently isn't.

b-rowan:
> **Copying the impl to every Braiins version** — I'd like to check something first. Right now only `v26_04` has the real implementation; `v21_09`/`v25_03`/`v25_05`/`v25_07` carry empty impls. Copying it is easy, but each also needs a `ConfigField::Timezone` location and a `send_graphql_command` call, which today only `v26_04` uses. Do the older BOS releases actually serve `bos { setTimezone }` and `timezone`/`timezoneList`? I can only test against BraiinsOS on one machine here, so if I copy it blind I'd be declaring `supports_timezone_config() == true` on four backends I can't verify — which is the sort of thing that reads as supported and then silently isn't.

This is what the `meta/` directory is for (https://github.com/256foundation/asic-rs/tree/master/meta/braiins)

The graphql schema should be pretty consistent across versions, but you can check against that OR the openapi schema, but they should all work in theory with the same implementation, this specific surface hasn't really changed.

pos-ei-don:
Pushed the Braiins copy in `2982385` — CI is green (Cargo Test, Cargo Assist, Conventional Commits, Python Test).

### What I checked before copying

You pointed me at `meta/`, so I went through the schemas there rather than copying blind:

| version | schema in `meta/braiins/` | `setTimezone` | `timezoneList` |
|---|---|:--:|:--:|
| 21.09 | `schema.graphql` | — | — |
| 24.09 | `schema.graphql` | ✔ | ✔ |
| 25.03 | `schema.graphql` | ✔ | ✔ |
| 25.05.1 | `schema.graphql` | ✔ | ✔ |
| 25.07 | `openapi.json` | n/a | n/a |
| 26.04 | `openapi.json` | n/a | n/a |

You were right that the surface is stable wherever it exists — with two wrinkles:

- **21.09 doesn't have it at all.** Neither the mutation nor the list is in its schema, so I left `v21_09` on the empty impl. Declaring `supports_timezone_config() == true` there would report support the firmware doesn't have.
- **25.07 and 26.04 only ship `openapi.json` in `meta/`,** which doesn't describe the GraphQL surface — you can see that from 26.04 scoring zero there while actually carrying the working implementation. So the schema can't confirm 25.07 either way. I copied it anyway on the strength of 26.04 working, but I'd rather label that inferred than verified. If you'd prefer `v25_07` stay empty until someone can test it against real hardware, say the word and I'll pull it back out.

Net: `v25_03` and `v25_05` are schema-covered, `v25_07` is inferred, `v21_09` stays empty. The impl in all three is byte-identical to the `v26_04` one — same query, same parse, same mutation.

### Still needs your call: the timezone type

This is the last open point, and I don't want to guess at it because it changes the public type.

Recap: `String` here isn't laziness, it's the common denominator. BraiinsOS gives real IANA zones via `timezoneList`; VNish has no named zones at all — it stores a fixed UTC offset as `"GMT+2"`, doesn't track DST, and has no list endpoint. So `chrono-tz` would type the Braiins side nicely and leave VNish with nowhere to go.

If you want the type safety anyway, the honest shape is probably an enum rather than one string:

```rust
pub enum Timezone {
    Named(String),   // IANA, e.g. "Europe/Vienna" — BraiinsOS
    Offset(i8),      // fixed UTC offset — VNish
}
```

so the two can't be confused and each backend declares which it speaks. Happy to build that, or to leave `String` as-is — either is fine by me, I just don't want to pick a public type on your behalf.

b-rowan:
> 25.07 and 26.04 only ship openapi.json in meta/, which doesn't describe the GraphQL surface — you can see that from 26.04 scoring zero there while actually carrying the working implementation. So the schema can't confirm 25.07 either way. I copied it anyway on the strength of 26.04 working, but I'd rather label that inferred than verified. If you'd prefer v25_07 stay empty until someone can test it against real hardware, say the word and I'll pull it back out.

This is fine, GraphQL was sort of "deprecated", they don't work on it anymore AFAIK, but the UI still uses it, so it should just be backwards compatible.

b-rowan:
> Recap: String here isn't laziness, it's the common denominator. BraiinsOS gives real IANA zones via timezoneList; VNish has no named zones at all — it stores a fixed UTC offset as "GMT+2", doesn't track DST, and has no list endpoint. So chrono-tz would type the Braiins side nicely and leave VNish with nowhere to go.

I think you might have my idea a bit confused here.  I'm saying "what if we use a pre-existing schema to represent the time zones inside asic-rs, so a user has a consistent format to pass into the functions", and we handle conversion.  I imagine something like chrono has timezone objects (haven't looked), which we can convert to and from "GMT-7" EG for VNish.  This means that the user always knows what they are getting, and we handle the weird abstraction to any miner specific types.

pos-ei-don:
Got it — I was answering a different question, sorry. One canonical type, we do the conversion, caller never sees the difference. Agreed.

I built it so there's something concrete to argue with, on a separate branch rather than pushed onto this PR: [`pos-ei-don/asic-rs@feat-timezone-canonical`](https://github.com/pos-ei-don/asic-rs/tree/feat-timezone-canonical). Compiles, clippy/fmt clean, 15 tests green.

What decides the type is that the conversion isn't symmetric. Braiins gives named IANA zones with DST rules, VNish stores a fixed offset (`"GMT+2"`, no DST, no list endpoint). Named → fixed drops the DST rule, so a miner set from `Europe/Vienna` would never shift back in winter. Fixed → named is ambiguous — plenty of zones sit on +02:00.

`chrono_tz::Tz` still works if VNish is restricted to the `Etc/GMT*` zones: real IANA zones, fixed offset, no DST rules, so both firmwares fit one type and round-trip cleanly. `available` becomes the 27 `Etc/GMT*` entries for VNish instead of the hardcoded strings I had.

Watch the sign — `Etc/GMT+2` is UTC−2, so VNish `"GMT+2"` maps to `Etc/GMT-2`. Wrong way round puts a miner four hours off and looks fine in review, so there's a test asserting the actual offset in seconds rather than just the names.

Two calls I made that you may want different:

- `TimezoneConfig.timezone` stays a `String` (the IANA name); the zone is reached through `TimezoneConfig::new(Tz)` / `from_tz` / `tz()`, and backends only ever touch `Tz`. An `Option<Tz>` field would need a newtype plus changes in `asic-rs-pydantic(-macros)`, which hardcodes the type→getter/schema mappings. Happy to go the newtype route if you'd rather have the field itself typed.
- `set_timezone_config` on the Python side now returns `bool` and raises instead of swallowing errors into `None`, otherwise the VNish rejection below is invisible from Python. Same shape as #332.

One behaviour question: `Europe/Vienna` on a VNish miner currently errors and names the `Etc/GMT*` equivalent for the offset in effect, rather than quietly pinning it to today's offset — that works all summer and is wrong in November. Say the word if you'd rather it pin.

And `chrono-tz` is a new dependency in `asic-rs-core` (it pulls `phf`). It's in the branch, easy to drop if you'd rather not take it — that one's yours to call.

pos-ei-don:
Pushed the implementation rather than describing it — easier to argue with a diff. CI is green.

The reasoning, briefly: the conversion isn't symmetric. Braiins gives named IANA zones with DST rules, VNish stores a fixed offset (`"GMT+2"`, no DST, no list endpoint). Named → fixed drops the DST rule, fixed → named is ambiguous. So `chrono_tz::Tz` works as one canonical type only if VNish is restricted to the `Etc/GMT*` zones — real IANA zones, fixed offset, no DST rules. Both firmwares then round-trip cleanly, and `available` is the same type everywhere.

One trap worth knowing about: `Etc/GMT+2` is UTC−2, so VNish `"GMT+2"` maps to `Etc/GMT-2`. There's a test asserting the actual offset in seconds rather than just the names, so a consistently inverted mapping can't pass.

This is a proposal, not a decision — any of it can change:

- **`chrono-tz` is a new dependency** in `asic-rs-core` (it pulls `phf`). Your call entirely; if you'd rather not take it, say so and I'll look for another way.
- `TimezoneConfig.timezone` stays a `String`, with the zone reached through `new(Tz)` / `from_tz` / `tz()`. A real `Option<Tz>` field would need a newtype plus changes in `asic-rs-pydantic(-macros)`, which hardcodes the type→getter/schema mappings. Happy to do that instead if you want the field itself typed.
- `Europe/Vienna` on a VNish miner errors and names the `Etc/GMT*` equivalent, rather than pinning to whatever offset applies today. Tell me if you'd rather it pin.
- `set_timezone_config` now raises from Python instead of returning `None` — otherwise that error is invisible there.

b-rowan:
> TimezoneConfig.timezone stays a String, with the zone reached through new(Tz) / from_tz / tz(). A real Option<Tz> field would need a newtype plus changes in asic-rs-pydantic(-macros), which hardcodes the type→getter/schema mappings. Happy to do that instead if you want the field itself typed.

As far as I know, chrono-tz is actually compatible with pyo3 directly.  If you enable the `chrono-tz` feature on pyo3, we can use it directly.  See https://docs.rs/pyo3/latest/pyo3/chrono_tz/index.html

If you do that, we should be able to use it directly rather than string, and avoid all the intermediate conversion.

pos-ei-don:
Done — you were right, pyo3's `chrono-tz` feature handles it. `TimezoneConfig.timezone` is `Option<Tz>` now and `available` is `Vec<Tz>`; the accessors and the `parse_iana` helper that only existed to bridge the string are gone. No pyo3 bump needed either — 0.28 already carries chrono-tz `>=0.10, <0.11` as an optional dependency and we're pinned at 0.10.4. CI is green, Python tests included.

Two small things that came with it:

- `asic-rs-pydantic` implements `PyPydanticType for Tz` along the `MacAddr` / `IpAddr` lines — schema `str`, accepts either a `ZoneInfo` or an IANA name, dumps the name. The macro's hint table needed an entry too (`Tz` → `tzinfo | str`); without one it falls through to `object`. No serializer-table entry, since `Tz` has a native `IntoPyObject` and a `String` mapping would put the hop straight back.
- Worth knowing rather than fixing: the getters hand out `zoneinfo.ZoneInfo(name)`, so a Python without tzdata for that key raises where a plain string never did. Fine on CI runners, just not free.

The VNish `GMT±N` ↔ `Etc/GMT∓N` mapping, the DST rejection and the Python raise are unchanged.

asic-rs-core/src/config/timezone.rs
```diff
@@ -0,0 +1,353 @@
...
+            "Timezone {tz} has a UTC{current} offset, which VNish cannot store \
+             (whole hours between UTC-12 and UTC+14 only)."
+        ),
+    }
+}
+
+/// [`tz_to_vnish_offset`] evaluated at the current instant.
+pub fn tz_to_vnish_offset_now(tz: Tz) -> anyhow::Result<String> {
+    tz_to_vnish_offset(tz, Utc::now())
+}
```
b-rowan: Move all VNish specific (or firmware specific functions) into the related firmware crate.  I would like to keep from coupling firmware to core directly like this, and have all the implementations for a specific firmware live in the same place.

asic-rs-core/src/config/timezone.rs
```diff
@@ -0,0 +1,353 @@
...
+        Ok(())
+    }
+
+    /// VNish `GMT-5` means UTC-5, which is `Etc/GMT+5` in POSIX terms.
+    #[test]
+    fn vnish_negative_offset_maps_to_etc_gmt_plus() -> anyhow::Result<()> {
+        assert_eq!(vnish_offset_to_tz("GMT-5")?, Tz::Etc__GMTPlus5);
+        assert_eq!(vnish_offset_to_tz("GMT-12")?, Tz::Etc__GMTPlus12);
+        Ok(())
+    }
```
b-rowan: Same with tests for VNish.

b-rowan (changes requested):
Very close now, just want to decouple core functionality from the VNish firmware stuff.  Could inline those functions into the `parse` function for VNish, or separate them out if you want to keep the tests.

pos-ei-don:
Heads-up: I'm off on holiday from today, back in about three weeks.

I'd wanted to get this over the line before leaving — the last rounds went quickly and your review made the code better each time. But I'd rather not rush the decoupling on the way out the door.

Where it stands: everything else is done and CI is green. What's left is exactly your last comment, and it turns out to be a clean cut — `vnish_offset_to_tz`, `tz_to_vnish_offset(_now)`, `fixed_offset_hours`, `tz_from_offset_hours` and the private helpers under them (`utc_offset_at`, `noon_utc`, `etc_gmt_equivalent`, `describe_offset`) are used only by the two VNish backends; no Braiins backend touches any of them. So the whole block plus its tests can move into the vnish crate, and `config/timezone.rs` keeps just the type.

Entirely your call from here: if you'd rather make that move yourself and merge, please go ahead, no need to wait for me. If you'd rather it sits until I'm back, that's fine too — I'll pick it up then.

PR #296, opened 2026-06-23
fix(python): improve the type hinting outputs of the pydantic macros

b-rowan:


asic-rs-pydantic/src/lib.rs
```diff
@@ -186,31 +186,13 @@ impl PyPydanticType for MacAddr {
...
-            return Ok(Self::from_secs_f64(seconds));
-        }
-        if let Ok(dict) = value.cast::<PyDict>() {
-            let secs = required_dict_item(dict, "secs")?.extract::<u64>()?;
-            return Ok(Self::from_secs(secs));
-        }
-        Err(PyValueError::new_err(
-            "Expected duration as timedelta, non-negative seconds, or {secs} dict",
-        ))
+        value.extract()
```
aleksander-rudolf, line 195: Gutting from_pydantic to value.extract() makes Duration rely entirely on field-path coercion. The generated model_validate classmethod (asic-rs-pydantic-macros/src/lib.rs:340-357) does not run the core schema — for a non-instance it calls from_pydantic(obj) directly. So a direct MinerData.model_validate({"uptime": 64493, ...}) (raw int/float/ISO-string seconds) will now fail, where 0.7.0's from_pydantic accepted int/float seconds. Every other type keeps parsing raw inputs in its own from_pydantic; Duration is now the exception that only works when something else coerces first.

b-rowan: Updated commit to also add a test for parsing from timedelta.  IMO we should parse the valid python output type in expected formats, in this case, timedelta and ISO-str.  I am removing the int/float parsing since those aren't valid timedelta types.

aleksander-rudolf (approved):

asic-rs-pydantic/src/lib.rs
jameshilliard: This still looks unfixed in the merged version. The added tests cover `MinerDataModel.model_validate(...)`, where Pydantic `timedelta_schema` coerces the field before Rust sees it, but the generated direct classmethod path still calls `<Self as PyPydanticType>::from_pydantic(obj)` directly. So `MinerData.model_validate({..., "uptime": 1.25})`, `64493`, or an ISO duration string still reaches `Duration::from_pydantic` as a raw Python value and fails because this implementation now only accepts an actual `timedelta`.

PR #301, opened 2026-06-29
fix(is-mining): ensure `is_mining` represents a paused state rather than hashrate

### LINKED ISSUE #303: discussion: is_mining representation
is_mining should represent whether the miner has been paused via software.

b-rowan:
`is_mining` should represent whether the miner is paused or not, rather than just checking if hashrate is 0.

A good example of this is for a watchdog, the watchdog should check the hashrate of the miner, and if it is bad, it should restart it, UNLESS the miner is paused by the user.


cfilipescu:
you forgot epic firmware. it has the same logic

PR #305, opened 2026-06-29
docs: add miner support review guidelines

cfilipescu:
## Summary
- add review guidelines for new miner and firmware support PRs
- document parser structure, hardware metadata, live API validation, supported controls, and live-test expectations
- include guidance for avoiding one-off abstractions and documenting live endpoint quirks

## Verification
- not run; docs-only change

### DIFF
--- a/docs/development/miner-support-review-guidelines.md
+++ b/docs/development/miner-support-review-guidelines.md
@@ -20,7 +20,7 @@ for behavior that cannot be validated with static fixtures.
 7. Implement supported controls if live endpoints exist; return unsupported only after checking.
 8. Verify live API method, endpoint, payload, auth, and response parsing instead of assuming browser or Antminer behavior.
 9. Add ignored live tests for destructive or hardware-dependent actions.
-10. Promote reusable cross-firmware behavior to core or shared utilities, not backend-local helpers.
+10. Keep backend-specific behavior self-contained unless a shared data model or core contract requires otherwise.
 
 ## Backend Structure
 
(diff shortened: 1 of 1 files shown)


docs/development/miner-support-review-guidelines.md
```diff
@@ -0,0 +1,154 @@
...
+1. Keep command constants local to the function that uses them.
+2. Match the structure of existing mature backends before inventing a new shape.
+3. Prefer early exits over deeply chained iterator parsing for miner data/config.
+4. Inline small one-off parsing helpers when they only obscure nearby logic.
+5. Use model hardware metadata for expected fans, boards, and chips when it is available.
+6. Avoid fixed magic ranges for boards or chains when the response or model can drive iteration.
+7. Implement supported controls if live endpoints exist; return unsupported only after checking.
+8. Verify live API method, endpoint, payload, auth, and response parsing instead of assuming browser or Antminer behavior.
+9. Add ignored live tests for destructive or hardware-dependent actions.
+10. Promote reusable cross-firmware behavior to core or shared utilities, not backend-local helpers.
```
b-rowan: Not entirely sure we want to do this?  I would prefer to silo implementations as much as possible, my general thought is that backends should be fully self contained as much as possible.  This isn't possible with the data model for example, but the less we have to share the better.

docs/development/miner-support-review-guidelines.md
```diff
@@ -0,0 +1,154 @@
...
+`value_by_keys` is not worth keeping if it only saves a few lines and makes the
+actual accepted response shape harder to see.
+
+Do not create a separate parse function and then perform similar parsing right
+after it. Keep the full parsing behavior in one place unless it is reused by
+multiple public paths or multiple firmware versions.
+
+For hashboards, avoid fixed upper bounds like `1..=9` unless the protocol truly
+requires that range. Prefer deriving chain indexes from response keys such as
+`chain_acnN`, or use `device_info.hardware.board_count()` when model metadata is
```
b-rowan: chain_acnN isn't a constant across multiple types, I would just remove this.

b-rowan (changes requested):

b-rowan (approved):

PR #308, opened 2026-07-01
fix: epic volcminer control board parsing

cfilipescu:
## Summary
- parse EPic PowerPlay `platform: TVXilinx` as VolcMiner `TVXilinx`
- keep generic AntMiner `Xilinx` platform parsing unchanged for AntMiner models
- handle generic Xilinx fallback as VolcMiner only when the detected make is VolcMiner
- add regression coverage for the ambiguous Xilinx cases

## Tests
- cargo test -p asic-rs-makes-volcminer
- MINER_IP=10.208.0.242 cargo test parse_data_live_test -p asic-rs-firmwares-volcminer -- --ignored --nocapture
- MINER_IP=192.168.13.235 cargo test parse_data_live_test_auto_detect -p asic-rs-firmwares-epic -- --ignored --nocapture

### DIFF
--- a/asic-rs-firmwares/epic/src/backends/v1/mod.rs
+++ b/asic-rs-firmwares/epic/src/backends/v1/mod.rs
@@ -1616,7 +1616,6 @@ mod tests {
         traits::firmware::MinerFirmware,
     };
     use asic_rs_makes_antminer::models::AntMinerModel;
-    use asic_rs_makes_volcminer::models::VolcMinerModel;
 
     use super::*;
     use crate::test::json::v1::{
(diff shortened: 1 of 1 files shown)


asic-rs-firmwares/epic/src/backends/v1/mod.rs
```diff
@@ -488,6 +489,19 @@ impl GetControlBoardVersion for PowerPlayV1 {
             return Some(cb.into());
         }
 
+        if let Some(platform) =
+            data.extract_nested::<String>(DataField::ControlBoardVersion, "platform")
+        {
+            let is_volcminer_platform = platform.trim().eq_ignore_ascii_case("TVXilinx");
+            let is_volcminer_model = self.device_info.make.eq_ignore_ascii_case("VolcMiner");
```
b-rowan: Better to compare against `VolcMinerMake::to_string` right?

asic-rs-firmwares/epic/src/backends/v1/mod.rs
```diff
@@ -488,6 +489,19 @@ impl GetControlBoardVersion for PowerPlayV1 {
             return Some(cb.into());
         }
 
+        if let Some(platform) =
+            data.extract_nested::<String>(DataField::ControlBoardVersion, "platform")
+        {
+            let is_volcminer_platform = platform.trim().eq_ignore_ascii_case("TVXilinx");
```
b-rowan, line 497: I don't think this check makes sense, we are already checking make, only case here is if its unknown.

asic-rs-firmwares/epic/src/backends/v1/mod.rs
```diff
@@ -504,6 +519,12 @@ impl GetControlBoardVersion for PowerPlayV1 {
                 Some(AntMinerControlBoard::BeagleBoneBlack).map(|cb| cb.into())
             }
             s if s.to_uppercase().contains("XILINX") => {
+                if self.device_info.make.eq_ignore_ascii_case("VolcMiner")
```
b-rowan: Same thing, compare against the type itself.

b-rowan (changes requested):

cfilipescu:
Updated the make checks to compare against `VolcMinerMake::default().to_string()` instead of the raw `"VolcMiner"` string.

I kept the explicit `platform == "TVXilinx"` guard because the live PowerPlay response that exposed this had `make=Unknown model=Unknown: undefined` with `platform: "TVXilinx"` and `cpu: "Xilinx Zynq Platform"`. If we only key off make there, that device falls through to the generic AntMiner Xilinx fallback again.

asic-rs-firmwares/epic/src/backends/v1/mod.rs
cfilipescu: Addressed in `a0bb667` by comparing against `VolcMinerMake::default().to_string()` instead of a raw make string.

asic-rs-firmwares/epic/src/backends/v1/mod.rs
cfilipescu: I think this one still needs to stay. The live response that triggered the bug reports `make=Unknown model=Unknown: undefined`, but the same payload has `platform: "TVXilinx"` and `cpu: "Xilinx Zynq Platform"`. If we only gate on make, this falls through to AntMiner `Xilinx` again. The `TVXilinx` guard is intentionally for that unknown-model PowerPlay case.

asic-rs-firmwares/epic/src/backends/v1/mod.rs
cfilipescu: Addressed in `a0bb667` as well. The fallback Xilinx branch now uses the same `VolcMinerMake::default().to_string()`-based `is_volcminer_model` value.

PR #315, opened 2026-07-17
fix: map AxeOS VR temp to board, chip temp to chain

adamdecaf:
AxeOS exposes distinct sensors on `/api/system/info`: `temp` (ASIC chip) and `vrTemp` (voltage regulator / board). Bitaxe and Nerdaxe backends copied `vrTemp` into inlet/outlet chip temperatures and often left `chips[].temperature` empty by reading `temp` from the config-only `/api/system/asic` payload.

This change is limited to the Bitaxe/Nerdaxe firmware backends:

- Map `vrTemp` → `board_temperature`
- Map `temp` → `inlet_chip_temperature` / `outlet_chip_temperature` and `chips[0].temperature` (preferring live system/info)
- Keep chip rows gated on `DataField::Chips`

No core data structure changes.

### DIFF
--- a/asic-rs-firmwares/bitaxe/src/backends/v2_0_0/mod.rs
+++ b/asic-rs-firmwares/bitaxe/src/backends/v2_0_0/mod.rs
@@ -280,12 +280,24 @@ impl GetHashboards for Bitaxe200 {
                 }
                 .as_unit(HashRateUnit::default())
             });
+        // `vrTemp` is the VR/board sensor; `temp` is the ASIC chip sensor.
+        let chip_temp = api_data
+            .get("temp")
+            .and_then(|v| v.as_f64())
+            .filter(|t| *t > -50.0)
+            .or_else(|| {
+                chip_data
+                    .and_then(|c| c.get("temp"))
+                    .and_then(|v| v.as_f64())
+                    .filter(|t| *t > -50.0)
+            })
+            .map(Temperature::from_celsius);
         board.board_temperature = api_data
             .get("vrTemp")
             .and_then(|v| v.as_f64())
             .map(Temperature::from_celsius);
-        board.inlet_chip_temperature = board.board_temperature;
-        board.outlet_chip_temperature = board.board_temperature;
+        board.inlet_chip_temperature = chip_temp;
+        board.outlet_chip_temperature = chip_temp;
         board.working_chips = api_data
             .get("asicCount")
             .and_then(|v| v.as_u64())
@@ -298,13 +310,10 @@ impl GetHashboards for Bitaxe200 {
             .get("frequency")
             .and_then(|v| v.as_f64())
             .map(Frequency::from_megahertz);
-        if let Some(chip_data) = chip_data {
+        if chip_data.is_some() {
             board.chips = vec![ChipData {
                 position: 0,
-                temperature: chip_data
-                    .get("temp")
-                    .and_then(|v| v.as_f64())
-                    .map(Temperature::from_celsius),
+                temperature: chip_temp,
                 voltage: board.voltage,
                 frequency: board.frequency,
                 tuned: Some(true),
@@ -608,6 +617,24 @@ mod tests {
         let hashboards_without_chips = miner.parse_hashboards(&data);
         assert!(hashboards_without_chips[0].chips.is_empty());
         assert_eq!(hashboards_without_chips[0].working_chips, Some(1));
+        assert_eq!(
+            hashboards_without_chips[0]
+                .board_temperature
+                .map(|t| t.as_celsius()),
+            Some(78.0)
+        );
+        assert_eq!(
+            hashboards_without_chips[0]
+                .inlet_chip_temperature
+                .map(|t| t.as_celsius()),
+            Some(59.875)
+        );
+        assert_eq!(
+            hashboards_without_chips[0]
+                .outlet_chip_temperature
+                .map(|t| t.as_celsius()),
+            Some(59.875)
+        );
 
         let mut collector = DataCollector::new_with_client(&miner, &mock_api);
         let data = collector
@@ -622,6 +649,12 @@ mod tests {
         assert_eq!(
             hashboards_without_chips[0].working_chips,
             hashboards_with_chips[0].working_chips
+        );
+        assert_eq!(
+            hashboards_with_chips[0].chips[0]
+                .temperature
+                .map(|t| t.as_celsius()),
+            Some(59.875)
         );
 
         let mut collector = DataCollector::new_with_client(&miner, &mock_api);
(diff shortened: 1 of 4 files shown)


asic-rs-core/src/data/board.rs
```diff
@@ -48,15 +48,18 @@ pub struct BoardData {
     pub hashrate: Option<HashRate>,
     /// The expected or factory hashrate of the board
     pub expected_hashrate: Option<HashRate>,
-    /// The board temperature, also sometimes called PCB temperature
+    /// The board temperature, also sometimes called PCB temperature.
+    /// On Bitaxe/Nerdaxe this is the VR sensor (`vrTemp`), not the ASIC chip.
```
b-rowan: No reason to include miner specific information in the core lib, remove.

asic-rs-core/src/data/board.rs
```diff
@@ -48,15 +48,18 @@ pub struct BoardData {
...
     #[serde(serialize_with = "serialize_temperature")]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub board_temperature: Option<Temperature>,
-    /// The temperature of the coolest / first-in-chain chip on the board
+    /// The temperature of the coolest / first-in-chain chip on the board.
+    /// On Bitaxe/Nerdaxe this is the ASIC chip sensor (`temp`), not VR.
```
b-rowan: Remove.

asic-rs-core/src/data/board.rs
```diff
@@ -48,15 +48,18 @@ pub struct BoardData {
...
     #[serde(serialize_with = "serialize_temperature")]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub inlet_chip_temperature: Option<Temperature>,
-    /// The temperature of the hottest / last-in-chain chip on the board
+    /// The temperature of the hottest / last-in-chain chip on the board.
+    /// On Bitaxe/Nerdaxe this is the ASIC chip sensor (`temp`), not VR.
```
b-rowan: Remove.

asic-rs-core/src/traits/miner.rs
```diff
@@ -267,14 +267,45 @@ impl<
...
+            } else if !chain_temps.is_empty() {
+                chain_temps
+            } else {
+                board_temps
+            };
+
+            if !temps.is_empty() {
                 Some(Temperature::from_celsius(
-                    board_temps.iter().sum::<f64>() / board_temps.len() as f64,
+                    temps.iter().sum::<f64>() / temps.len() as f64,
```
b-rowan: Core functionality like this is intentional, but could probably be labeled better.  Maybe change the documentation on the MinerData field to explicitly note this is average board temperature?

b-rowan (changes requested):
Concept ACK, but I think this should be limited to fixes specifically on the relevant miner types, and not touch the core data structures (other than the small documentation changed mentioned if you feel it's required).

adamdecaf:
That's my bad. Changing core types wasn't my intention - was going a bit too fast with Grok.

b-rowan:
Still a couple core changes, just revert them fully.  Making average be chip OR board is even more confusing...

asic-rs-firmwares/bitaxe/src/backends/v2_0_0/mod.rs
```diff
@@ -298,20 +312,17 @@ impl GetHashboards for Bitaxe200 {
...
+        // from system/info `temp`, not the config-only ASIC endpoint.
+        board.chips = vec![ChipData {
+            position: 0,
+            temperature: chip_temp,
+            voltage: board.voltage,
+            frequency: board.frequency,
+            tuned: Some(true),
+            working: Some(true),
+            hashrate: board.hashrate.clone(),
+        }];
```
b-rowan: I believe the reason the check for chip data was here is for guarding for `DataField::Chips`, since having that check will skip adding the chips if it is excluded by the user.

```suggestion
        if let Some(chip_data) = chip_data {
	        board.chips = vec![ChipData {
	            position: 0,
	            temperature: chip_temp,
	            voltage: board.voltage,
	            frequency: board.frequency,
	            tuned: Some(true),
	            working: Some(true),
	            hashrate: board.hashrate.clone(),
	        }];
	    }
```

asic-rs-firmwares/bitaxe/src/backends/v2_0_0/mod.rs
```diff
@@ -606,8 +617,42 @@ mod tests {
...
+        );
+        // Guard against the old bug that copied VR into chip-chain fields.
+        assert_ne!(
+            hashboards_without_chips[0]
+                .board_temperature
+                .map(|t| t.as_celsius()),
+            hashboards_without_chips[0]
+                .inlet_chip_temperature
+                .map(|t| t.as_celsius())
+        );
```
b-rowan: Can revert these test changes with the fix above, these should have no chips since they should be explicitly excluded by the user.

asic-rs-firmwares/nerdaxe/src/backends/v1/mod.rs
```diff
@@ -295,20 +309,15 @@ impl GetHashboards for NerdAxeV1 {
...
-        }
+        board.chips = vec![ChipData {
+            position: 0,
+            temperature: chip_temp,
+            voltage: board.voltage,
+            frequency: board.frequency,
+            tuned: Some(true),
+            working: Some(true),
+            hashrate: board.hashrate.clone(),
+        }];
```
b-rowan: Save thing here, can re-add the check.


```suggestion
        if let Some(chip_data) = chip_data {
            board.chips = vec![ChipData {
                position: 0,
                temperature: chip_temp,
                voltage: board.voltage,
                frequency: board.frequency,
                tuned: Some(true),
                working: Some(true),
                hashrate: board.hashrate.clone(),
            }];
        }
```

b-rowan (changes requested):
One more small thing I noticed, other than that this should be good.

adamdecaf:
Addressed the remaining feedback:

- Restored the `DataField::Chips` guard so chip rows are only emitted when chips are requested
- Reverted the related test expectations (`chips` stays empty when excluded)
- Rebased onto latest `master` as a single commit; no core changes

asic-rs-firmwares/bitaxe/src/backends/v2_0_0/mod.rs
adamdecaf: Done — restored the DataField::Chips guard.

asic-rs-firmwares/bitaxe/src/backends/v2_0_0/mod.rs
adamdecaf: Reverted; chips stay empty when DataField::Chips is excluded. Kept only board/inlet/outlet mapping assertions.

asic-rs-firmwares/nerdaxe/src/backends/v1/mod.rs
adamdecaf: Done — same guard restored here.

b-rowan (approved):
LGTM 🚀

PR #320, opened 2026-07-29
fix: improve large subnet scan reliability

### LINKED ISSUE #319: Improve large-network scanning with staged TCP probing
### Problem

  asic-rs scans become unreliable under high concurrency. On a /22 network:

  - Nmap found 478 hosts with ports 80 or 4028 open in 18 seconds.
  - asic-rs found 461 miners with concurrency 64 and a 10-second timeout, but took 9 minutes.
  - Higher concurrency was faster but missed many devices.
  - Retrying misses at lower concurrency recovered most devices.

  Address concurrency can also produce multiple simultaneous HTTP/RPC requests per host, increasing the actual network load.

  ### Proposal

  Use a two-stage scan:

  1. Concurrently probe miner TCP ports with a short timeout and dedicated concurrency limit.
  2. Run application-level identification only on responsive hosts, using a lower independent concurrency limit.

  The scanner should distinguish between:

  - Identified miners.
  - TCP-responsive but unidentified hosts.
  - Unresponsive hosts.

  The existing port check is not ideal for this because it checks ports sequentially and acts as a hard gate.

Before selecting defaults, I can benchmark probe timeout, probe concurrency, identification concurrency, and retries across repeated scans and select some statistically supported values to minimize time to detect all miners on the large network through a vpn (which is a common use case) 

---

 Would this design fit MinerFactory, or is another separate scanner abstraction preferred?

DanNicolau:
Closes #319 

## Summary

  - Fixed connectivity retry semantics.
  - Probe all four ports concurrently rather than sequentially.
  - Complete TCP probing across the range before starting identification.
  - Use separate TCP and identification concurrency limits.
  - Correct connectivity_retries, which previously acted more like a boolean gate and did not perform the configured number of
    attempts.
  - Added separate TCP/identification concurrency, timeout, retry, and strict-gating controls.
  - Set measured defaults: TCP 256/500 ms, identification 128/3 s, construction 5 s, no retries.
  - Probe common miner ports concurrently and stream identifications as they finish.
  - Added HTTP request deadlines and a miner-construction deadline.
  - Added matching Rust/Python builder APIs and documentation.

 ## Performance

11.7s to scan a /22 subnet with 441 target devices

  ## Reliability considerations

  The initial benchmark ran Nmap immediately before each test, which warmed neighbor and ARP state. Cold end-to-end scans demonstrated that
  TCP probing could still miss responsive miners transiently. For this reason, TCP results are not a hard exclusion by default.

  Strict filtering is still available for callers that prefer minimum scan time and can tolerate possible false negatives.

  ## Validation

  - cargo test -p asic-rs
  - All asic-rs doctests
  - cargo clippy -p asic-rs --all-targets --all-features -- -D warnings
  - Python builder methods and type stubs updated
  - End-to-end validation on the target /22 network


b-rowan:
50 seconds to scan a /22 seems really high to me?  Gotta be some way to improve past that...

DanNicolau:
Yeah, I'm not thrilled about that result either (it's closer to 2m with the tcp probe and the miner discovery added together). I'll move it from a draft once I get something I think is effective.

b-rowan:
Not sure if its just me or what, scanning a local network, /20 subnet (~4k addresses), it seems to have no consistency issues AFTER an initial run.  For some reason that first run just wants to miss stuff.  Is there anything we can do to make the ARP table less garbage at its job?

DanNicolau:
I also do see significant improvements once the ARP table/cache starts getting filled out.

b-rowan:
Some sloppified suggestions, maybe somewhere to start looking.   I have curated out the irrelevant ones, but some things to look into here possibly.


---

## Scan reliability and concurrency findings

After reviewing `factory.rs`, the relevant issues are:

### Confirmed issues

1. **`connectivity_retries` does not perform retries**

   The current check only determines whether the configured value is greater than one. Regardless of the configured value, each port is probed at most once.

   This can cause false negatives during transient packet loss, initial ARP resolution, or network congestion.

2. **Sequential port checks can be slow, while parallel checks could amplify concurrency**

   Ports `80`, `4028`, `4029`, and `8889` are checked sequentially, with an early return after the first successful connection.

   The early-return behavior is intentional, but on an unreachable or filtered host, each timeout is incurred in sequence. With the default one-second timeout, a host can take roughly four seconds to reject.

   Simply checking all ports concurrently would improve latency but could multiply the number of simultaneous connections by four. A globally bounded or staggered port race may provide a better balance.

### Possible contributing issues

3. **The configured concurrency limit applies to hosts, not individual connections**

   `buffer_unordered` limits the number of hosts being scanned, but each responsive host can then execute multiple discovery commands concurrently.

   As a result, the actual number of active sockets may be several times higher than the configured host concurrency. The existing file-descriptor estimate accounts for some fan-out, but it does not enforce a hard socket limit.

   Possible changes include:

   * Reducing the default host concurrency.
   * Adding a separate global connection semaphore.
   * Deriving the host limit from the number of expected connections per host.

4. **There is no connection-start rate limit**

   A large concurrency value allows many connections—and potentially many ARP requests—to begin in a short burst.

   Even when the scanner has sufficient file descriptors, this can overload network queues, neighbor discovery, firewalls, or embedded devices and lead to intermittent false negatives.

   A connection-rate limiter may improve accuracy without substantially reducing throughput.

### Lower-priority follow-ups

5. **Discovery configuration is rebuilt for each responsive host**

   The firmware registry and deduplicated discovery-command set could be calculated once per scan and shared between host scans. This would reduce allocations and repeated setup work, though it is unlikely to be the main cause of the current scanning problem.

6. **HTTP discovery responses are read without an explicit size limit**

   Reading the complete response is useful for some device identification methods, but unrelated web services could return large responses. A configurable or sufficiently generous response limit may be worth considering.

## Suggested initial changes

The highest-priority change is to fix `connectivity_retries` so it performs the configured number of attempts.

After that, testing should compare:

* Lower host concurrency.
* A separate global socket limit.
* Staggered or bounded concurrent port checks.
* A connection-start rate limit.

Metrics for timeouts, refused connections, unreachable hosts, resource errors, and retry recoveries would help determine whether missed devices are caused by transient network loss or local resource exhaustion.

b-rowan:
Maybe also worth looking at the relevant scanning section from btctools, it seems pretty consistent.

https://github.com/btccom/libbtctools/blob/master/src/miner/MinerScanner.cpp

DanNicolau:
I'll take a look, I appreciate the direction.

DanNicolau:
Benchmarks significantly improved with these defaults, and changes to the tcp retries. Sometimes on a cold network scan it will drop considerable miners, but I don't see a simple way to get around that. In this test environment over a vpn to another farm the ARP caching happens downstream so just running it again is the best I can do.

I found that there was a potential unlimited wait for certain umcos endpoints for the identification stage, and those now time out. I was unable to reproduce those situations running repeated scans for 6 hours but it's the only place I see the potential for multi-minute scans to get stuck.

Probably best for other issues so I didn't include the changes here:
- global semaphore -- right now because we parallelize the tcp-probe ports (which has an observable benefit rather than serializing all the ports) we can spawn more connections than our tcp concurrency. to be more correct we would probably need to take connections from a global semaphore.
- connection start rate limiting -- I couldn't isolate this from ARP table warming up downstream so it may or may not have a benefit. It is probably also dependent on downstream networking hardware.

---

## Update summary

  ### Updated defaults

  - TCP probe concurrency: 256 addresses
  - Firmware-identification concurrency: 128 hosts
  - TCP probe timeout: 500 ms
  - TCP retries: 0 additional attempts
  - Identification retries: 1 additional attempt
  - Identification retry concurrency: 8
  - Identification retry backoff: 250 ms
  - Strict TCP filtering enabled by default
  - Firmware-identification timeout: 3 seconds
  - Miner-construction timeout: 5 seconds
  - Discovery HTTP connect timeout: 2 seconds
  - Discovery HTTP total timeout: 3 seconds

  TCP ports 80, 4028, 4029, and 8889 are checked concurrently for each address. TCP probing and firmware identification remain separate bounded stages.

  The connectivity retry behavior was corrected so that configuring zero retries still performs the initial TCP attempt.

  APIs.

  ### Benchmark results

  Testing used a routed /22 network with approximately 461 supported devices.

  The selected configuration consistently found 460 devices in five consecutive end-to-end release scans, completing in approximately 19.6–21.5 seconds.

  Identification concurrency remained above 99% recall through 256, but degraded at higher values:

   Concurrency    Mean detected    Minimum recall    Mean duration
  ━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━
           128            460.0            99.78%           20.5 s
  ─────────────  ───────────────  ────────────────  ───────────────
           256            460.0            99.78%           20.8 s
  ─────────────  ───────────────  ────────────────  ───────────────
           320            459.3            99.57%           36.7 s
  ─────────────  ───────────────  ────────────────  ───────────────
           384            457.0            97.83%           22.3 s

  A separate TCP-probe benchmark produced 100% recall across five runs at concurrency 256 and averaged approximately 1.58 seconds. Misses began appearing at concurrency 478.

  Latency measurements across roughly 4,500 successful device constructions supported the new timeout bounds. Miner construction had a p99.9 of approximately 2.7 seconds and a maximum of approximately 3.0 seconds, leaving meaningful
  headroom beneath the five-second deadline.

  A global socket semaphore and connection-start rate limiter were not added. The current evidence did not isolate connection-start rate as a problem independently of network and neighbour-cache warm-up effects.

  ### Validation

  - 9 unit tests passed
  - 10 doctests passed
  - Strict Clippy passed with warnings denied
  - Formatting and diff checks passed
  - Repeated end-to-end validation completed on the target /22 network

DanNicolau:
I'd also like some input on the desired scope of PRs. I can subdivide PRs into more atomic changes like update constants, make tcp port scan parallel etc if that would make review easier. In my head these are all scan improvements so they belong together.

These current changes are full of AI slop so I still have to review them before I take this out of draft -- but I am happy with what the benchmarking shows.

src/factory.rs
```diff
@@ -29,19 +29,23 @@ use ipnet::IpNet;
...
-const IDENTIFICATION_TIMEOUT: Duration = Duration::from_secs(10);
-const CONNECTIVITY_TIMEOUT: Duration = Duration::from_secs(1);
-const CONNECTIVITY_RETRIES: u32 = 3;
+const IDENTIFICATION_TIMEOUT: Duration = Duration::from_secs(3);
+const MINER_CONSTRUCTION_TIMEOUT: Duration = Duration::from_secs(5);
+const IDENTIFICATION_CONCURRENCY: usize = 128;
+const CONNECTIVITY_TIMEOUT: Duration = Duration::from_millis(500);
+const CONNECTIVITY_RETRIES: u32 = 0;
+const CONNECTIVITY_CONCURRENCY: usize = 256;
+const CONNECTIVITY_RETRY_CONCURRENCY: usize = 64;
```
b-rowan, line 38: I feel like these should use the same semaphore as the base on right?  Basically just scope the semaphore around the full loop w/ retries?

src/factory.rs
```diff
@@ -29,19 +29,23 @@ use ipnet::IpNet;
...
+const CONNECTIVITY_RETRY_BACKOFF: Duration = Duration::from_millis(100);
+const SCAN_RETRIES: u32 = 0;
+const SCAN_RETRY_BACKOFF: Duration = Duration::from_millis(250);
+const IDENTIFICATION_RETRY_CONCURRENCY: usize = 8;
```
b-rowan, line 42: Same as above, scope the semaphore to include the retry loop.

src/factory.rs
```diff
@@ -29,19 +29,23 @@ use ipnet::IpNet;
...
 
 fn calculate_optimal_concurrency(ip_count: usize) -> usize {
-    match ip_count {
-        0..=1000 => 1000,
-        1001..=5000 => 2500,
-        5001..=10000 => 5000,
-        _ => 10000,
-    }
+    ip_count.clamp(1, IDENTIFICATION_CONCURRENCY)
 }
```
b-rowan, line 49: Can probably drop this if using a constant value?  Just use clamp inline if needed.

src/factory.rs
```diff
@@ -50,6 +54,10 @@ fn calculate_desired_nofile_limit(concurrency: usize) -> u64 {
         .max(MIN_NOFILE_LIMIT)
 }
 
+fn calculate_retry_concurrency(concurrency: usize) -> usize {
+    concurrency.clamp(1, IDENTIFICATION_RETRY_CONCURRENCY)
+}
```
b-rowan, line 59: Same thing here, fine to inline this.

src/factory.rs
```diff
@@ -60,6 +68,40 @@ async fn check_port_open(ip: IpAddr, port: u16, connectivity_timeout: Duration)
...
+async fn probe_ip_batch(
+    ips: Vec<IpAddr>,
+    connectivity_timeout: Duration,
+    concurrency: usize,
+) -> (Vec<IpAddr>, Vec<IpAddr>) {
+    let mut reachable = Vec::new();
+    let mut unreachable = Vec::new();
+    let mut probes = stream::iter(ips)
+        .map(|ip| async move { (ip, check_miner_ports(ip, connectivity_timeout).await) })
+        .buffer_unordered(concurrency.max(1));
```
b-rowan, line 93: Hardcoded max 1 concurrent here?

src/factory.rs
```diff
@@ -278,25 +365,26 @@ impl Default for MinerFactory {
...
+            if attempt > 0 {
+                let multiplier = 1_u32 << (attempt - 1).min(10);
+                tokio::time::sleep(self.connectivity_retry_backoff.saturating_mul(multiplier))
+                    .await;
             }
-            if check_port_open(ip, 8889, self.connectivity_timeout).await {
+            if check_miner_ports(ip, self.connectivity_timeout).await {
                 return self.get_miner(ip).await;
             }
         }
```
b-rowan, line 381: Could wrap this whole check in the factory global semaphore?

src/factory.rs
```diff
@@ -688,6 +888,120 @@ impl MinerFactory {
...
+                    Arc::clone(&factory),
+                    unreachable,
+                    identification_concurrency,
+                ));
+                while let Some(outcome) = fallback.next().await {
+                    yield outcome;
+                }
+            }
+        })
+    }
```
b-rowan, line 1003: Pretty complex function with stacking the retries in this way, I really feel like retries should be consolidated with the initial request, just to aggregate some of the different variables and flatten this out a bit.

b-rowan (changes requested):
Concept seems fine, but I think we just end up falling back to more tuning of default parameters.  The novel part here to me is introducing the backoff logic and fixing some places with missing timeouts, but I would like to try to consolidate some of those variables if possible.  Tried to note places where this seems possible, let me know if you have any questions...

DanNicolau:
Closing this and will be superceded by more focused PRs

PR #325, opened 2026-08-12
fix(antminer): read power draw from the `new_api` stats payload

cryptographicturk:
> **Depends on #328.** That PR fixes the RPC transport so `new_api` actually
> reaches the miner. Until it merges, its commit appears here too; afterwards
> this PR is the single `mod.rs` change. Reworked from the web endpoint to RPC
> per review.

`wattage` is always `None` on newer stock firmware. `DataField::Wattage` is
sourced from the legacy RPC `stats` payload, which carries no power reading on
these generations.

## Where the reading lives

The `new_api` variant of `stats` reports it. That is a different payload to the
legacy one — its `STATS` array holds a single element, so the value sits at
`/STATS/0` rather than the `/STATS/1` the legacy source uses:

```
{"command":"stats"}                  -> STATS len 2, Msg "CGMiner stats", no power
{"command":"stats","new_api":true}   -> STATS len 1, Msg "stats",         power/watt present
```

## The change

- Adds `stats` + `new_api` as a second source for `DataField::Wattage`, read at
  `/STATS/0`.
- Accepts the `watt` key alongside the existing `power` / `Power` /
  `chain_power`. The spelling varies by model on identical firmware: the L9
  reports `power`, the L11 `watt`.

The legacy location is kept and tried first, so firmware that does report power
there is unaffected. Applied to both `v2020` and `v2023_07`.

## Measured

Against live hardware, all previously `None`:

```
L11   3651 W        L9   3357 W
L11   3636 W        L9   3353 W
                    L9   3379 W
```

An idle unit reports its true low draw rather than falling back to absent, so a
genuine zero stays distinguishable from a missing reading.

No regressions in the groups that cannot benefit: T21, S21 Hydro and S21+ Hydro
expose no power draw on any transport and continue to report nothing, and older
units that do not honour `new_api` still receive the legacy payload exactly as
before.

The two payloads share only `fan_num` and `rate_30m`, identical in both, so
merging them into one field cannot silently clobber a value.

## Verification

- `cargo clippy -p asic-rs-firmwares-antminer --all-targets` — clean
- `cargo test -p asic-rs-firmwares-antminer` — 20 passed
- `cargo test --workspace` — no failures
- Live check against multiple L9 and L11 units

### DIFF
--- a/asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
+++ b/asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
@@ -509,14 +509,37 @@ impl GetDataLocations for AntMinerV2020 {
                     tag: None,
                 },
             )],
-            DataField::Wattage => vec![(
-                RPC_STATS,
-                DataExtractor {
-                    func: get_by_pointer,
-                    key: Some("/STATS/1"),
-                    tag: None,
-                },
-            )],
+            DataField::Wattage => {
+                // Newer stock firmware reports power draw only in the `new_api`
+                // variant of `stats`. That is a different payload to the legacy
+                // one: its `STATS` array holds a single element, so the reading
+                // sits at `/STATS/0` rather than the `/STATS/1` used above.
+                //
+                // Not a `const` because `json!` is not const-evaluable.
+                let rpc_new_stats = MinerCommand::RPC {
+                    command: "stats",
+                    parameters: Some(json!({ "new_api": true })),
+                };
+
+                vec![
+                    (
+                        RPC_STATS,
+                        DataExtractor {
+                            func: get_by_pointer,
+                            key: Some("/STATS/1"),
+                            tag: None,
+                        },
+                    ),
+                    (
+                        rpc_new_stats,
+                        DataExtractor {
+                            func: get_by_pointer,
+                            key: Some("/STATS/0"),
+                            tag: None,
+                        },
+                    ),
+                ]
+            }
             DataField::SerialNumber => vec![
                 (
                     WEB_SYSTEM_INFO,
@@ -835,6 +858,9 @@ impl GetWattage for AntMinerV2020 {
             if let Some(power) = stats_data
                 .get("power")
                 .or_else(|| stats_data.get("Power"))
+                // Same firmware version spells this differently per model: the
+                // L9 reports `power`, the L11 `watt`.
+                .or_else(|| stats_data.get("watt"))
                 .and_then(|v| v.as_f64())
             {
                 return Some(Power::from_watts(power));
--- a/asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
+++ b/asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
@@ -510,11 +510,6 @@ impl GetDataLocations for AntMinerV2020 {
                 },
             )],
             DataField::Wattage => {
-                // Newer stock firmware reports power draw only in the `new_api`
-                // variant of `stats`. That is a different payload to the legacy
-                // one: its `STATS` array holds a single element, so the reading
-                // sits at `/STATS/0` rather than the `/STATS/1` used above.
-                //
                 // Not a `const` because `json!` is not const-evaluable.
                 let rpc_new_stats = MinerCommand::RPC {
                     command: "stats",
--- a/asic-rs-firmwares/antminer/src/backends/v2020/rpc.rs
+++ b/asic-rs-firmwares/antminer/src/backends/v2020/rpc.rs
@@ -10,33 +10,33 @@ use asic_rs_core::{
 use async_trait::async_trait;
 use serde_json::{Value, json};
 
+/// Bitmain's one API extension to the cgminer protocol.
+const NEW_API_FLAG: &str = "new_api";
+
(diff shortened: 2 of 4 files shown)


asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
```diff
@@ -391,6 +391,11 @@ impl GetDataLocations for AntMinerV2020 {
             parameters: None,
         };
 
+        const WEB_STATS: MinerCommand = MinerCommand::WebAPI {
+            command: "stats",
+            parameters: None,
+        };
```
b-rowan: AFAIK you can actually use the RPC endpoint to get this output, the source of this file just calls the cgminer API with `new_api=True`, so this should actually work:


```suggestion
        let rpc_new_stats: MinerCommand = MinerCommand::RPC {
            command: "stats",
            parameters: Some(json!({ "new_api": true })),
        };
```

b-rowan (changes requested):
I think using the RPC endpoint is preferable (it should be faster than the web and doesn't need auth).

asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
cryptographicturk: Thanks — I tried this, and it doesn't reach the data. I think it points at a
  real bug though.

  `send_rpc_command` (`v2020/rpc.rs:31`) nests `parameters` under cgminer's
  `"parameter"` key. The firmware doesn't recognise `new_api` there — it silently
  answers with the legacy payload. Tested against three units on `86.48-2.0.0`:

  ```
  model  request form                                      result
  L9     {"command":"stats","parameter":{"new_api":true}}  STATS[2], "CGMiner stats", power ABSENT
  L9     {"command":"stats","new_api":true}                STATS[1], "stats",         power = 3414
  L11    {"command":"stats","parameter":{"new_api":true}}  STATS[2], "CGMiner stats", power ABSENT
  L11    {"command":"stats","new_api":true}                STATS[1], "stats",         watt  = 3648
  L11    {"command":"stats","parameter":{"new_api":true}}  STATS[2], "CGMiner stats", power ABSENT
  L11    {"command":"stats","new_api":true}                STATS[1], "stats",         watt  = 3638
  ```

  The nested form returns a response byte-identical to plain `{"command":"stats"}`
  — the parameter is simply ignored. I built the suggested change and ran it
  against real hardware: `wattage` went to `None` on every L9 and L11. The failure
  is silent (`STATUS: "S"`, well-formed body), which is why it passes `cargo check`
  and the unit tests and only shows up against a live miner.

  The wider consequence: `stats`, `summary`, `pools`, `rate`, `warning` and
  `reload` (`rpc.rs:65–107`) all pass `Some(json!({"new_api": true}))` through
  that same path, so none of them is actually enabling `new_api` today.

  Fixing it properly is a transport change with a wider blast radius than this PR:
  `send_command` forwards arbitrary parameters, and genuine cgminer `parameter`
  usage (`switchpool` etc.) takes a string. One workable rule would be "object
  parameters merge at top level, scalars stay under `parameter`", which repairs
  all six helpers without breaking the legitimate cases. Happy to do that as a
  separate PR and rebase this onto RPC afterwards if you'd prefer it that way —
  I have L9/L11 hardware to verify against.

  For now I've kept the web endpoint and corrected the comment, which previously
  claimed the RPC payload carried no power reading at all. That was wrong, and it
  was the basis for your suggestion.

  Two notes on scope while I'm here:

  - The `watt` key handling isn't incidental — on identical firmware the L9
    reports `power` and the L11 reports `watt`, so both spellings are needed
    whichever transport ends up being used.
  - I've only verified this successfully on L9 and L11.

cryptographicturk:
I updated it after submitting another PR to fix an underlying issue that prevented this solution from working.

asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
```diff
@@ -509,14 +509,37 @@ impl GetDataLocations for AntMinerV2020 {
...
-                    tag: None,
-                },
-            )],
+            DataField::Wattage => {
+                // Newer stock firmware reports power draw only in the `new_api`
+                // variant of `stats`. That is a different payload to the legacy
+                // one: its `STATS` array holds a single element, so the reading
+                // sits at `/STATS/0` rather than the `/STATS/1` used above.
+                //
+                // Not a `const` because `json!` is not const-evaluable.
```
b-rowan: Full comment not needed, this is sortof implied by the flag.  Non const comment is fine.

asic-rs-firmwares/antminer/src/backends/v2020/rpc.rs
```diff
@@ -10,6 +10,35 @@ use asic_rs_core::{
...
+        }
+        Some(params) => json!({
+            "command": command,
+            "parameter": params
+        }),
+        None => json!({
+            "command": command
+        }),
+    }
+}
```
b-rowan, line 40: See comments on #328

b-rowan (changes requested):
Nits only, just shorten a comment.  Contingent on #328, small fixes there then I will merge both.

b-rowan (approved):

PR #328, opened 2026-08-13
fix(antminer): send `new_api` as a top-level flag rather than a `parameter`

cryptographicturk:
`send_rpc_command` wraps every parameter value in cgminer's `parameter`
argument. Bitmain's `new_api` is not a `parameter` — it is a top-level flag
alongside `command`. Nested, the firmware ignores it and answers with the
legacy payload.

Every `new_api` helper on this client is affected — `stats`, `summary`,
`pools`, `rate`, `warning` and `reload` all ask for the new API and silently
receive the old one. Nothing errors: the response is a well-formed
`STATUS: "S"` body, just the wrong shape, so a caller cannot tell.

## Evidence

Against an L9 and an L11 on `86.48-2.0.0`:

```
{"command":"stats","new_api":true}                -> STATS len 1, Msg "stats",         power/watt present
{"command":"stats","parameter":{"new_api":true}}  -> STATS len 2, Msg "CGMiner stats", no power field
```

The second response is byte-identical to sending `{"command":"stats"}` with no
parameter at all.

## The change

Object parameters merge at the top level; scalars keep the `parameter`
wrapper, preserving cgminer's own convention for commands like `switchpool`
that take a pool index. Every parameter currently passed within this backend
is a `new_api` object, so nothing relied on the old wrapping.

Request construction moves into `build_rpc_request` so the wire format is
unit-testable without a socket.

## Verification

- `cargo clippy -p asic-rs-firmwares-antminer --all-targets` — clean
- `cargo test -p asic-rs-firmwares-antminer` — 20 passed (8 new, covering
  object merge, scalar wrapping, absent parameters, and `command` precedence
  over a conflicting key)
- `cargo test --workspace` — no failures

Checked against live hardware by temporarily sourcing `DataField::Wattage`
from `stats` + `new_api` (that wiring is *not* part of this PR). With the fix,
wattage resolves on hardware that reports it and stays absent elsewhere:

```
L11   3651 W      L9   3357 W
L11   3636 W      L9   3353 W
                  L9   3379 W
```

Models whose firmware exposes no power draw (T21, S21 Hydro, S21+ Hydro)
correctly report nothing rather than erroring, and older units that do not
honour `new_api` at all still receive the legacy payload exactly as before.

The two payloads share only `fan_num` and `rate_30m`, identical in both, so
merging them into one field carries no risk of silently clobbering a value.

### DIFF
--- a/asic-rs-firmwares/antminer/src/backends/v2020/rpc.rs
+++ b/asic-rs-firmwares/antminer/src/backends/v2020/rpc.rs
@@ -10,33 +10,33 @@ use asic_rs_core::{
 use async_trait::async_trait;
 use serde_json::{Value, json};
 
+/// Bitmain's one API extension to the cgminer protocol.
+const NEW_API_FLAG: &str = "new_api";
+
 /// Build the JSON request for a cgminer-style RPC call.
 ///
-/// Bitmain's API extensions — `new_api` being the one this backend uses — are
-/// top-level flags alongside `command`, not values of cgminer's `parameter`
-/// argument. Wrapping them makes the firmware ignore the flag and silently
-/// answer with the legacy payload, which is indistinguishable from a
-/// successful call.
-///
-/// An object is always one of those extensions; cgminer's own convention
-/// passes a scalar (e.g. `switchpool`'s pool index), so scalars keep the
-/// `parameter` wrapper.
+/// `new_api` is a top-level flag alongside `command`, not a value of cgminer's
+/// `parameter` argument. Nested under `parameter` the firmware ignores it and
+/// answers with the legacy payload — a well-formed success a caller cannot
+/// tell apart from the real thing. Everything else keeps the `parameter`
+/// wrapper cgminer expects, e.g. `switchpool`'s pool index.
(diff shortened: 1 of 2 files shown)


asic-rs-firmwares/antminer/src/backends/v2020/rpc.rs
```diff
@@ -10,6 +10,35 @@ use asic_rs_core::{
...
+        }
+        Some(params) => json!({
+            "command": command,
+            "parameter": params
+        }),
+        None => json!({
+            "command": command
+        }),
+    }
+}
```
b-rowan, line 40: Is it preferable to only check for the `new_api` key specifically?  AFAIK this is the only key that needs to be top level, so might just be easiest to check for it directly.

If the need arises in the future, I think we would consider adding a `flags` argument to the rpc command, which would add those top level flags.

b-rowan (changes requested):
I think we should just check for the new_api flag directly, rather than implementing something generic (yet).  See my comment.

b-rowan (approved):

PR #329, opened 2026-08-18
feat(types): derive TypeScript bindings for shared models

Erisli:
## Summary
- Add ts-rs as a workspace dependency and configure large integers as TypeScript numbers.
- Derive TS for shared config, telemetry, model, hardware, pool, message, and make-specific enums.
- Override TypeScript shapes for custom-serialized measurement, MAC, power target, and uptime fields.

## Validation
- cargo fmt
- cargo +1.95.0 check --workspace

Note: cargo check --workspace with the local 1.88.0 override hit an existing compiler limitation in asic-rs-core/src/traits/miner.rs:419, so validation used the already-installed 1.95.0 toolchain.

### DIFF
--- a/asic-rs-core/src/config/fan.rs
+++ b/asic-rs-core/src/config/fan.rs
@@ -1,9 +1,10 @@
 #[cfg(feature = "python")]
 use pyo3::prelude::*;
 use serde::{Deserialize, Serialize};
+use ts_rs::TS;
 
 #[cfg_attr(feature = "python", pyclass(skip_from_py_object, module = "asic_rs"))]
-#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, TS)]
 /// Fan control mode.
 pub enum FanMode {
     /// Firmware-managed fan control using a target temperature.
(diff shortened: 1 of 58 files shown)


b-rowan:
Is this dependent on #322 ?

asic-rs-core/src/config/fan.rs
```diff
@@ -13,7 +14,7 @@ pub enum FanMode {
 }
 
 #[cfg_attr(feature = "python", pyclass(skip_from_py_object, module = "asic_rs"))]
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, TS)]
```
b-rowan, line 17: Probably a good idea to feature gate these, same as the python derives.  That will depend on #322.

b-rowan (changes requested):
Concept is good, just need to merge the actual bindings then I would like the derives feature-gated.

asic-rs-core/src/config/fan.rs
Erisli: got it! is there any futher action needed from me for this pr before #332 is merged?

cfilipescu (approved):

b-rowan (approved):
Talked with @cfilipescu, going to defer feature gating for now, we will work on better javascript bindings shortly.  #322 is a bit weird right now AFAIK, but WIP on making that better.

PR #330, opened 2026-08-19
test(epic): add live perpetual tuning test

cfilipescu:
## Summary

- add an ignored ePIC PowerPlay live test that reads and prints the current perpetual tuning and scaling configuration before writing
- support ChipTune and PowerTune selection through `EPIC_TUNE_MODE`
- support optional `EPIC_HASHRATE_TARGET_TH` and `EPIC_POWER_TARGET_W` overrides, defaulting to 103 TH/s and 3000 W
- keep the general live data test read-only by removing its tuning write

## Why

This provides a deliberate, opt-in way to validate perpetual tuning against a live miner while preserving the miner's existing scaling values and keeping ordinary test runs non-destructive.

## Validation

- `cargo +1.95.0 fmt --all -- --check`
- `cargo +1.95.0 test -p asic-rs-firmwares-epic` (9 passed, 3 live tests ignored)

### DIFF
--- a/asic-rs-firmwares/epic/src/backends/v1/mod.rs
+++ b/asic-rs-firmwares/epic/src/backends/v1/mod.rs
@@ -1910,13 +1910,6 @@ mod tests {
         );
 
         println!(
-            "set_tuning_config {}",
-            miner
-                .set_tuning_config(tuning_config, Some(scaling_config))
-                .await?
-        );
-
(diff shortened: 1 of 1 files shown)


b-rowan (approved):
Looks good, was gonna talk about the commit message but looks like thats fine now.

PR #335, opened 2026-08-24
fix(hashrate): read algorithm and unit from the miner instead of assuming SHA-256

cryptographicturk:
## Problem

Scrypt Antminers (L3+/L7/L9/L11) come back as **SHA-256 miners** under both VNish and stock firmware, and every hashrate is read on the assumption that it is in GH/s.

None of this depended on what the miner replied — the values were compile-time literals, so no device response could change the outcome.

## Three causes

**1. `AntMinerModel` never overrode `MinerModel::hash_algorithm()`**, so it inherited the trait's SHA-256 default. Only VolcMiner and Elphapex override it, and both return `Scrypt` unconditionally because every model in those makes is Scrypt. AntMiner is the one genuinely mixed-algorithm make in the repo — SHA-256, Scrypt, X11, Kadena, Equihash, Ethash, Handshake — and the one that never used the hook.

**2. Both firmwares hardcoded the algorithm** — `HashAlgorithm::SHA256` into `DeviceInfo::new`, `algo: "SHA256"` into every `HashRate`. Five sites per VNish backend, three per stock backend.

**3. Both hardcoded the unit, although both firmwares report it.** VNish publishes `hr_measure` (`GH/s` / `MH/s` / `N/A`) on `/info` — the same response already parsed for the model — documented in the 1.2.x and 1.3.x API specs under `meta/`. Stock cgminer publishes `rate_unit` in STATS; it is sitting next to `total_rateideal` in the repo's own S21 fixture (`rate_unit: "GH"`).

## Approach

The algorithm now comes from the model; the unit comes from the miner.

For VNish, `hr_measure` is carried alongside each hashrate figure using the collector's existing `tag` mechanism (the same pattern whatsminer/proto/luxminer already use), so **no core API change** was needed.

For stock, the two sources are deliberately kept separate: cgminer's SUMMARY key names its own unit (`GHS 5s` vs `MHS 5s`), so that key is trusted for the value it labels, while `rate_unit` is used only for the STATS figures — `total_rateideal`, `chain_rate*` — whose names say nothing about their scale. Crossing the two would let an inconsistent `rate_unit` corrupt a value read from a correctly-named key.

## Non-regression

Everything falls back to GH/s when the unit is absent or unusable (`"N/A"`), and `GHS 5s` is tried first, so SHA-256 models take exactly the path they did before. The pre-existing `test_antminer` fixture tests still pass unchanged, which is the signal that matters.

## Scope left out

Models whose algorithm `HashAlgorithm` cannot yet name — KS3/KS5 (kHeavyHash), K7 (Eaglesong), E9 Pro (Ethash), Z15 (Equihash), HS3 (Handshake), DR5 (Blake256R14) — still report SHA-256. Unchanged from today, and documented in a comment rather than left silent. Naming them means adding public enum variants that flow into the Python and TypeScript bindings, which seemed like a separate call — happy to fold it in if you would prefer.

Also left out, as neither is this bug: VNish control-board parsing (the firmware sends short codes `aml|bb|cv|stm|xil`, and `AntMinerControlBoard::parse` only knows `AML`/`BB`, so `cv`/`xil`/`stm` yield no control board), and the VNish `set_power_limit` assumption that autotune presets are named by wattage.

## Not verified against hardware

That a Scrypt stock Antminer names its SUMMARY key `MHS 5s`. I could not confirm it, so it is only consulted as a fallback after `GHS 5s` misses — a wrong guess yields today's behaviour rather than something worse. If anyone has an L7/L9 on stock firmware, a `summary` + `stats` dump would settle it.

## Verification

Run on Rust 1.98.0 stable (`--workspace` needs Python dev headers for `asic-rs-pydantic` to link, so the three changed crates were targeted):

- `cargo test` — 33 passed, 0 failed
- `cargo clippy --all-targets` — clean, with `warnings = "deny"`

Seven new tests: `l_series_is_scrypt`, `d_series_is_x11`, `sha256_models_are_unchanged`, `unknown_model_defaults_to_sha256`, plus fixture-driven `scrypt_model_is_read_in_the_unit_it_reports` for stock v2020 and both VNish backends, and `sha256_model_is_unchanged` / `unusable_unit_falls_back_to_gigahash` guarding the fallback.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01ErTo1a4BgYptp5tC8r1oHR


asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
```diff
@@ -685,27 +715,48 @@ impl GetHashboards for AntMinerV2020 {
...
-        data.extract_map::<f64, _>(DataField::Hashrate, |f| {
+        let summary = data.get(&DataField::Hashrate)?;
+
+        // Preferred key first. `GHS 5s` is what every SHA-256 model reports, so
+        // those keep taking exactly the path they did before.
+        let (value, unit) = [
+            ("GHS 5s", HashRateUnit::GigaHash),
+            ("MHS 5s", HashRateUnit::MegaHash),
+            ("MHS av", HashRateUnit::MegaHash),
+        ]
```
b-rowan: Why not split on " " and then just send the first part to HashRateUnit directly?  Should be more extensible this way, and I have yet to see any example that doesn't follow this pattern...

asic-rs-firmwares/antminer/src/backends/v2023_07/mod.rs
```diff
@@ -661,27 +691,48 @@ impl GetHashboards for AntMinerV202307 {
...
-        data.extract_map::<f64, _>(DataField::Hashrate, |f| {
+        let summary = data.get(&DataField::Hashrate)?;
+
+        // Preferred key first. `GHS 5s` is what every SHA-256 model reports, so
+        // those keep taking exactly the path they did before.
+        let (value, unit) = [
+            ("GHS 5s", HashRateUnit::GigaHash),
+            ("MHS 5s", HashRateUnit::MegaHash),
+            ("MHS av", HashRateUnit::MegaHash),
+        ]
```
b-rowan: Same as previous comment.

asic-rs-firmwares/vnish/src/backends/v1_2_0/mod.rs
```diff
@@ -752,6 +793,18 @@ impl GetPools for VnishV120 {
...
+    /// every SHA-256 model reports — when it is absent or unusable, so this
+    /// cannot regress the models that worked before.
+    fn hashrate_unit(field: Option<&Value>) -> HashRateUnit {
+        field
+            .and_then(|v| v.pointer("/unit"))
+            .and_then(|v| v.as_str())
+            .and_then(|s| HashRateUnit::from_str(s).ok())
+            .unwrap_or(HashRateUnit::GigaHash)
+    }
+
```
b-rowan: Inline this instead, don't need a separate function definition, this is short and simple enough.

asic-rs-firmwares/vnish/src/backends/v1_3_0/mod.rs
```diff
@@ -764,6 +805,18 @@ impl GetPools for VnishV130 {
...
+    /// every SHA-256 model reports — when it is absent or unusable, so this
+    /// cannot regress the models that worked before.
+    fn hashrate_unit(field: Option<&Value>) -> HashRateUnit {
+        field
+            .and_then(|v| v.pointer("/unit"))
+            .and_then(|v| v.as_str())
+            .and_then(|s| HashRateUnit::from_str(s).ok())
+            .unwrap_or(HashRateUnit::GigaHash)
+    }
+
```
b-rowan: Inline.

asic-rs-makes/antminer/src/models.rs
```diff
@@ -149,6 +150,28 @@ impl MinerModel for AntMinerModel {
...
+    fn hash_algorithm(&self) -> HashAlgorithm {
+        match self {
+            Self::L3Plus | Self::L3PlusPlus | Self::L7 | Self::L9 | Self::L11 => {
+                HashAlgorithm::Scrypt
+            }
+            Self::D3 | Self::D7 | Self::D9 => HashAlgorithm::X11,
+            Self::KA3 => HashAlgorithm::Kadena,
+            _ => HashAlgorithm::SHA256,
+        }
+    }
```
b-rowan, line 184: Concept here is good, but this is not super extensible if a new model is added, it adds another place where date needs to be input.  Is there any better way to do this?

If there's nothing better this is fine, but want to take a second pass at this.

b-rowan (changes requested):
Couple small nits, main thing is just to double check the Antminer model stuff.  The other question to ask is if we want to more strictly define the algorithm types into an enum or something, then just implement string conversion, but that can likely be a seperate issue.

asic-rs-firmwares/antminer/src/backends/v2020/mod.rs
cryptographicturk: Done — good call. Now reads the key as `<unit> <window>` and hands the first part to `HashRateUnit::from_str`, preferring the `5s` window and falling back to `av`:

```rust
let (value, unit) = ["5s", "av"].into_iter().find_map(|window| {
    summary.iter().find_map(|(key, raw)| {
        let (unit, key_window) = key.split_once(' ')?;
        if key_window != window { return None; }
        Some((Self::parse_f64_field(raw)?, HashRateUnit::from_str(unit).ok()?))
    })
})?;
```

This also removed the one assumption in the PR I could not verify. The old version had to *know* that Scrypt models say `MHS 5s`; this one just reads whatever unit the key names, so `KHS`/`THS` work without a code change.

asic-rs-firmwares/antminer/src/backends/v2023_07/mod.rs
cryptographicturk: Same change applied here.

asic-rs-firmwares/vnish/src/backends/v1_2_0/mod.rs
cryptographicturk: Inlined at all three call sites, helper removed.

asic-rs-firmwares/vnish/src/backends/v1_3_0/mod.rs
cryptographicturk: Inlined here too.

asic-rs-makes/antminer/src/models.rs
cryptographicturk: Agreed, the `match` was a second place to keep in sync. The algorithm is now declared on the variant itself via strum's property:

```rust
#[serde(alias = "ANTMINER L9")]
#[strum(props(algo = "Scrypt"))]
L9,
```
```rust
fn hash_algorithm(&self) -> HashAlgorithm {
    self.get_str("algo")
        .and_then(|algo| HashAlgorithm::from_str(algo).ok())
        .unwrap_or(HashAlgorithm::SHA256)
}
```

Adding a model is one place now, and the value sits next to its siblings where it is hard to miss. An absent property means SHA-256, which covers most of the make. Since `HashAlgorithm` already derives `EnumString`, the string conversion is the derive rather than anything hand-written — which is also roughly the direction of your enum note on the main review.

Open to alternatives if you would rather not have the data in attributes; deriving it from the model name was the other option I considered, but `KA3` / `KS5` / `K7` are three different algorithms behind the same prefix, so it would not hold up.

b-rowan (approved):

PR #342, opened 2026-08-25
refactor(models): ensure every model has a valid algo property

b-rowan:


asic-rs-macros/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    "Blake2S256",
+    "Kadena",
+    "KHeavyHash",
+    "Eaglesong",
+    "EtHash",
+    "Equihash",
+    "Handshake",
+    "Blake256R14",
+    "Unknown",
+];
```
b-rowan, line 18: Not ideal, but can't really figure out a better way to do this.

cfilipescu (approved):

asic-rs-macros/src/lib.rs
cfilipescu: I think it's fine for now

PR #344, opened 2026-08-26
fix(factory): bound miner discovery operations

DanNicolau:
Added timeouts to discovery ops. The first of a few PRs I'm going to submit for scan improvements. Contributes to #319 
---
This change prevents miner discovery from hanging indefinitely.

  Previously, the identification timeout covered only the firmware-detection commands. Once firmware was identified,
  constructing the miner object could perform additional network requests without being covered by that deadline. HTTP
  connections also lacked an explicit connection timeout in some paths.

  The change:

  - Applies the existing identification timeout to the entire operation:
      - Run firmware-detection commands.
      - Select the detected firmware.
      - Construct the firmware-specific miner object.

  - Adds explicit connection and total-request deadlines to discovery HTTP clients.
  - Treats an expired discovery deadline as “miner not identified” (Ok(None)), allowing the rest of the scan to continue.
  - Adds tests for a stalled HTTP response and stalled miner construction.
  - Updates Rust/Python documentation to clarify the timeout’s end-to-end meaning.

  It deliberately does not change concurrency, retry behavior, port probing, scan staging, or defaults. Consequently, normal
  scans perform essentially the same; the benefit appears when a device or connection stalls.

### DIFF
--- a/asic-rs-core/src/util.rs
+++ b/asic-rs-core/src/util.rs
@@ -26,8 +26,18 @@ pub fn unix_timestamp_secs() -> u64 {
     )
 }
 
+/// Build an HTTP client with bounded connection and total-request deadlines
+/// for firmware discovery.
 pub fn build_discovery_client() -> Result<reqwest::Client, ModelSelectionError> {
+    build_discovery_client_with_timeout(DEFAULT_RPC_TIMEOUT)
+}
+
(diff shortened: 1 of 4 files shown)


b-rowan (approved):
LGTM

PR #348, opened 2026-08-31
feat(epic): support hashrate-split pool status

Erisli:
## Summary

Adds ePIC PowerPlay pool parsing for hashrate-split mode while preserving standard pool behavior.

## Changes

- Fetches and tags hashrate-split config/status data for pool collection.
- Parses split pool groups with quota, active pool state, share counts, and connection status.
- Aligns pool config parsing with the tagged hashrate-split config payload.

## Validation

- Reviewed diff against master.
- git diff --check passed.
- cargo test -p asic-rs-firmwares-epic parse_ currently stops in asic-rs-core/src/traits/miner.rs with existing lifetime bound error before backend tests run.

### DIFF
--- a/asic-rs-firmwares/epic/src/backends/v1/mod.rs
+++ b/asic-rs-firmwares/epic/src/backends/v1/mod.rs
@@ -98,6 +98,163 @@ impl PowerPlayV1 {
         pools
     }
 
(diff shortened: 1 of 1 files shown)


b-rowan (approved):
Looks good.  I tend to dislike breaking stuff out of the parsing functions (I find it easier to read inline), but these are such long functions I think it's fine here.

PR #355, opened 2026-09-09
fix(tuning): represent disabled ptune as manual target

### LINKED ISSUE #352: Expose perpetual tuning running status in MinerData
Consumers can read requested and scaled tuning targets, but cannot distinguish perpetual tuning being disabled from tuning telemetry being unavailable. Expose the firmware-reported running status as an optional typed field in `MinerData` (suggested name: `tuning_running: Option<bool>`).

### Existing source and gap

Checked at `560e72f12e300200db2c5b63762e3fb917200cf6`; current `master` `6ef32df811437a87b07594db19d118c8bf0940fe` has the same relevant behavior.

- [MinerData](https://github.com/256foundation/asic-rs/blob/560e72f12e300200db2c5b63762e3fb917200cf6/asic-rs-core/src/data/miner.rs#L128) exposes `tuning_target` and `scaled_tuning_target`, but no running flag.
- The EPIC/UMC backend already obtains the summary through [DataField::TuningTarget](https://github.com/256foundation/asic-rs/blob/560e72f12e300200db2c5b63762e3fb917200cf6/asic-rs-firmwares/epic/src/backends/v1/mod.rs#L425).
- [Target parsing](https://github.com/256foundation/asic-rs/blob/560e72f12e300200db2c5b63762e3fb917200cf6/asic-rs-firmwares/epic/src/backends/v1/mod.rs#L1164) checks `/PerpetualTune/Running` and returns no target when false or unavailable. Consumers cannot recover a reliable running flag from target presence.
- Optimization completion is separately available through `BoardData.tuned`; it has different semantics.

### Acceptance

- Preserve the explicit UMC running flag: true, false, or unknown.
- False requires an explicit false value; missing/null/malformed data and unsupported backends produce `None`.
- Document that this is firmware-reported perpetual/autotuning running status, independent of whether hashing is running or optimization has completed.
- Use other backends only where an equivalent reported status exists. Do not infer it from `is_mining`, target presence, or board tuning completion.
- Include the value in ordinary `get_data()` snapshots and the existing field/getter machinery as appropriate.
- Existing target/scaled-target telemetry remains available for requested-versus-effective displays such as `70(60) TH/s`; those targets already exist and are outside this field addition.

### Implementation and verification

Integrate with the existing collection and parsing path, reusing already-required responses. Preserve unsupported/missing values as `None`. Update the Rust model, serialization, Python models/stubs, generated TypeScript types where applicable, and documentation together.

Add fixture-based parser/serialization coverage in the existing CI-run tests. No live miner or device configuration changes should be needed.

Cover explicit true/false, absent/null/invalid flag, missing summary, running while optimization is incomplete, and a backend without support.

cfilipescu:
## Summary

- add a board-indexed `Manual` variant to `TuningTarget`, with all other variants implying perpetual tuning is enabled
- represent each board setpoint as optional frequency and voltage values so an explicit disabled state is retained when telemetry is incomplete
- serialize available manual frequencies and voltages as scalar MHz and volts, matching `BoardData`, with scalar round-trip deserialization
- return the manual target from ePIC telemetry and config parsing whenever `PerpetualTune.Running` is explicitly false
- preserve missing or malformed running status as `None` and expose the board-level representation through Python, Pydantic, TypeScript, and type stubs
- reject unsupported manual configuration in other firmware backends

## Testing

- `cargo test --workspace --all-features`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `pytest -q python/tests` (98 passed)

Closes #352


asic-rs-firmwares/epic/src/backends/v1/mod.rs
```diff
@@ -1933,6 +1964,119 @@ mod tests {
...
+                    { "Index": 0, "Data": 480 },
+                    { "Index": 1, "Data": 490 },
+                ]
+            }
+        });
+        let data = HashMap::from([(DataField::TuningTarget, summary.clone())]);
+        let expected = TuningTarget::Manual {
+            voltage: Some(Voltage::from_volts(12.6)),
+            frequency: Some(Frequency::from_megahertz(485.0)),
+        };
```
b-rowan, line 2000: This is a bit... weird...

Maybe `Manual` needs to have a tuning setting per board?

`Manual (Vec<(frequency, voltage)>)`?

b-rowan: Probably better to use a hashmap `{0:(freq, volt), 1:(freq,volt)}`

PR #357, opened 2026-09-10
feat(data): expose detailed firmware operating state

### LINKED ISSUE #353: Expose detailed operating state in MinerData
Expose an optional firmware-reported operating-state value in `MinerData` so consumers can display states such as `Mining`, `Idling`, and `AdjustingClockVoltage`, and group miners by those states.

### Existing source and gap

Checked at `560e72f12e300200db2c5b63762e3fb917200cf6`; current `master` `6ef32df811437a87b07594db19d118c8bf0940fe` has the same relevant behavior.

- [MinerData](https://github.com/256foundation/asic-rs/blob/560e72f12e300200db2c5b63762e3fb917200cf6/asic-rs-core/src/data/miner.rs#L139) exposes messages and `is_mining: bool`, but no detailed state.
- The [EPIC/UMC collector](https://github.com/256foundation/asic-rs/blob/560e72f12e300200db2c5b63762e3fb917200cf6/asic-rs-firmwares/epic/src/backends/v1/mod.rs#L527) already extracts `/Status/Operating State` into `DataField::IsMining`.
- [Boolean parsing](https://github.com/256foundation/asic-rs/blob/560e72f12e300200db2c5b63762e3fb917200cf6/asic-rs-firmwares/epic/src/backends/v1/mod.rs#L1231) reduces that string to `state != "Idling"` and defaults to true when absent. Detailed states are lost during normalization.

### Acceptance

- Add optional typed operating-state telemetry (suggested representation: `operating_state: Option<String>`) to the ordinary `MinerData` snapshot.
- Preserve UMC's actual reported state, including adjustment/startup/error states and unfamiliar future values.
- Missing/null/invalid state data produces `None`; it must not become an invented `Mining` or `Idling` state.
- A string representation or extensible typed representation must retain unknown firmware labels.
- Support additional backends when they provide explicit state information; keep detailed state unavailable when only a coarse boolean can be obtained.
- Preserve the existing `is_mining` API and document how the detailed state differs from that boolean.
- Consume the existing summary response; avoid duplicate requests.
- This issue covers operating state only. Existing last-error messages already have a separate representation; last command and error-parser enhancements are outside scope.

### Implementation and verification

Integrate with the existing collection and parsing path, reusing already-required responses. Preserve unsupported/missing values as `None`. Update the Rust model, serialization, Python models/stubs, generated TypeScript types where applicable, and documentation together.

Add fixture-based parser/serialization coverage in the existing CI-run tests. No live miner or device configuration changes should be needed.

Cover Mining, Idling, adjustment and unknown states, absent/null/invalid values, boolean-only firmware, and preservation through typed serialization/bindings.

cfilipescu:
Fixes #353.

## Summary

- Add optional `MinerData.operating_state` and `get_operating_state()` with a shared enum covering mining/stable operation, startup, tuning, frequency/voltage adjustment, pause/stop, degraded mining, and errors.
- Parse explicit state telemetry from ePIC/UMC, VNish, Braiins REST, MARA, and Proto. Preserve unfamiliar values as `Unknown { raw }`; missing, invalid, or boolean-only telemetry remains `None`.
- Reuse existing collector requests without changing legacy `is_mining` behavior or existing `DataField` numeric values.
- Update Python/Pydantic bindings and stubs, TypeScript types, documentation, and regression coverage. Backends without detailed state telemetry use the default unavailable result.

## Validation

- `cargo test --all --locked --exclude asic-rs-pydantic --exclude asic-rs-pydantic-macros`
- `cargo test --workspace --all-features --locked --exclude asic-rs-pydantic --exclude asic-rs-pydantic-macros`
- `cargo clippy --workspace --all-features --locked`
- `cargo fmt --all -- --check`
- Rebuilt the Python extension with Maturin; `python -m pytest -q python/tests`: **114 passed** on Python 3.13.
- `git diff --check`

Regression tests cover firmware-to-snapshot mappings, unknown/raw and missing states, Rust/Pydantic serialization, TypeScript shape, boolean-only firmware, and single-request ePIC summary collection.

Live-miner tests were not run.


asic-rs-firmwares/epic/src/backends/v1/mod.rs
```diff
@@ -524,7 +524,7 @@ impl GetDataLocations for PowerPlayV1 {
                     },
                 ),
             ],
-            DataField::IsMining => vec![(
+            DataField::IsMining | DataField::OperatingState => vec![(
```
b-rowan: I would like to keep these isolated, just improves readability when checking for something specific.

asic-rs-firmwares/marathon/src/backends/v1/mod.rs
```diff
@@ -553,7 +553,7 @@ impl GetDataLocations for MaraV1 {
                     tag: None,
                 },
             )],
-            DataField::IsMining => vec![(
+            DataField::IsMining | DataField::OperatingState => vec![(
```
b-rowan: Same here, fix in all these cases if possible.

PR #362, opened 2026-09-10
fix(core): use algorithm-specific hashrate units

### LINKED ISSUE #304: expand hashrate struct in core to implement default units for display
Might be worth breaking hashrate out to being a trait, then having a few different hashrate types (ScryptHashRate, SHA256HashRate, etc), which can each implement default_unit and display for example.

cfilipescu:
## Summary

- derive the conventional display unit from `HashRate.algo`
- normalize miner, expected, and hashboard rates through `as_default_unit()` across every backend
- infer the algorithm-specific unit in the Python constructor when no explicit unit is supplied
- expose default-unit lookup and conversion in the Rust and Python APIs

The defaults are:

| Algorithm | Unit |
| --- | --- |
| SHA-256, Blake2S256, Kadena, kHeavyHash, Eaglesong, Handshake, Blake256R14 | TH/s |
| Scrypt, X11 | GH/s |
| EtHash | MH/s |
| Equihash | KH/s |
| Unknown | H/s |

Explicitly supplied units are still preserved, `HashRateUnit::default()` remains TH/s for compatibility, and the serialized `HashRate` shape is unchanged.

## Testing

- `cargo test --workspace --all-targets`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo fmt --all -- --check`
- Python test suite: 114 passed
- targeted Antminer Scrypt parser tests

Closes #304


asic-rs-core/src/data/hashrate.rs
```diff
@@ -223,6 +223,28 @@ pub struct HashRate {
...
+            HashAlgorithm::SHA256
+            | HashAlgorithm::Blake2S256
+            | HashAlgorithm::Kadena
+            | HashAlgorithm::KHeavyHash
+            | HashAlgorithm::Eaglesong
+            | HashAlgorithm::Handshake
+            | HashAlgorithm::Blake256R14 => HashRateUnit::TeraHash,
+            HashAlgorithm::Unknown => HashRateUnit::Hash,
+        }
+    }
```
b-rowan: Probably best to put this match directly on the algo enum instead?  Then when defining a new algorithm it is in the same place and obvious you have to add a unit to match it?

asic-rs-core/src/data/hashrate.rs
```diff
@@ -318,3 +358,47 @@ impl Div<HashRate> for Power {
...
+            (HashAlgorithm::Unknown, HashRateUnit::Hash),
+        ] {
+            let hashrate = HashRate {
+                value: 1.0,
+                unit: HashRateUnit::Hash,
+                algo,
+            };
+            assert_eq!(hashrate.default_unit(), unit, "{algo}");
+        }
+    }
```
b-rowan: I don't think we need a test for a match statement, I assume rust has that figure out by now 😆

python/tests/test_pydantic_interop.py
```diff
@@ -485,6 +485,30 @@ def test_hashrate_constructor_accepts_unit_enum(
...
+def test_hashrate_constructor_uses_algorithm_default_unit(
+    algorithm: HashAlgorithm, default_unit: HashRateUnit
+) -> None:
+    constructed = HashRate(1.5, algo=algorithm)
+    explicit = HashRate(1.5, HashRateUnit.H, algorithm)
+
+    assert constructed.unit == default_unit
+    assert constructed.default_unit() == default_unit
+    assert explicit.unit == HashRateUnit.H
+    assert explicit.into_default_unit().unit == default_unit
```
b-rowan: Again not sure if this test is needed.

b-rowan (changes requested):

asic-rs-core/src/data/hashrate.rs
cfilipescu: Addressed in 72169c4. The match now lives on HashAlgorithm as default_hashrate_unit(), and HashRate delegates to it.

asic-rs-core/src/data/hashrate.rs
cfilipescu: Removed the redundant match-table test in 806c246. The behavior-focused Scrypt conversion test remains.

python/tests/test_pydantic_interop.py
cfilipescu: Removed the parametrized Python mapping test in 3195cf2.

b-rowan (approved):

PR #368, opened 2026-09-15
Feat: add get_miner_direct to skip port pre-check

nkatha23:
`MinerFactory::get_miner(ip)` runs a TCP port pre-check against ports 80, 4028, 4029, and 8889 before attempting firmware identification. This is the right default for bulk scanning, but the wrong behaviour when the caller already holds a live connection to the miner, imo. 

In pool software, miners dial the pool over Stratum. The pool has a live session and knows theminer's IP, but management ports are often firewalled for inbound probes. `get_miner` returns `Ok(None)`,  indistinguishable from the miner being
offline.

This PR Adds `MinerFactory::get_miner_direct(ip)`: same panic-catch and identification timeout as `get_miner`, but calls `get_miner_inner`directly without the port pre-check.

PS: `send_rpc_command` hardcodes port 4028, which causes port collisionsbetween parallel tests. Worked around with a static mutex for now.Making ports configurable per-factory can be worked on separately as a follow-up.


src/factory.rs
```diff
@@ -400,6 +400,34 @@ impl MinerFactory {
...
+            }
+            Err(_) => {
+                tracing::debug!(
+                    timeout_ms = self.identification_timeout.as_millis(),
+                    "miner discovery timed out for {ip}"
+                );
+                Ok(None)
+            }
+        }
+    }
```
b-rowan, line 429: This function is exactly identical to `get_miner` as it is right now in master...

See below:

https://github.com/256foundation/asic-rs/blob/476b1772099b4151784d7d2f61d932c981a81657/src/factory.rs#L381-L401

b-rowan (commented):
`get_miner_direct` is functionally identical to `get_miner` as it is right now?  Was this intentional?

src/factory.rs
nkatha23: Oh- okay. My bad. I missed that.

nkatha23: @b-rowan would you be open to me working on the hardcoded ports issue?

b-rowan: > @b-rowan would you be open to me working on the hardcoded ports issue?

No opposition, although I'm not sure in what situation this would occur?  Only reason I can think of is if you are trying to "spoof" a large amount of miners on a single server and have their API's accessible on different ports, like what was done for testing hydrapool+mujina?

Happy to accommodate if you can explain what this is needed for, but otherwise I am hesitant to add more complexity for no reason.  Feel free to open a related issue as well and we can discuss there.

nkatha23: > > @b-rowan would you be open to me working on the hardcoded ports issue?
> 
> No opposition, although I'm not sure in what situation this would occur? Only reason I can think of is if you are trying to "spoof" a large amount of miners on a single server and have their API's accessible on different ports, like what was done for testing hydrapool+mujina?
> 
> Happy to accommodate if you can explain what this is needed for, but otherwise I am hesitant to add more complexity for no reason. Feel free to open a related issue as well and we can discuss there.

The use case is pool software integrating asic-rs, like [braidpool](https://github.com/braidpool/braidpool/issues/501). In enterprise and hosted mining deployments, operators commonly remap management ports away from defaults (80, 4028) to avoid automated scanners. A miner can connect outbound to the pool on Stratum fine, but its management 
API lives on a non-standard port. Without configurable ports, asic-rs silently fails to identify it.

The hardcoded port also makes unit testing painful,  tests that exercise the RPC path have to compete for localhost:4028, which is what the mutex workaround in this PR was addressing. Configurable ports would let tests use TcpListener::bind(:0) and get OS-assigned ports, eliminating the collision entirely. I'll open an issue

PR #382, opened 2026-09-16
feat(discovery): add DiscoveryCommand with explicit port field

### LINKED ISSUE #370: feat: make RPC and HTTP management ports configurable per MinerFactory
send_rpc_command hardcodes port 4028, send_web_command hardcodes port 80. This surfaces as a  problem in two places today.

Test suite port collisions. Any test that exercises the RPC discovery path needs to bind localhost:4028. Two such tests running in parallel hit Address already in use (os error 98). PR #368 worked around this with a static tokio::sync::Mutex<()> to serialize those tests. As more tests are added this serialization bottleneck grows. If tests could request an OS-assigned port via TcpListener::bind(:0), the collision disappears entirely and tests can run in parallel without coordination.

Multi-miner hosts and remapped ports. When multiple miner processes run on a single server (the pattern used in hydrapool/mujina test infrastructure), each process needs a distinct port. asic-rs currently cannot address any instance other than the one on the default 4028/80. The same limitation affects production deployments where operators remap management ports away from defaults to reduce scanner exposure, or put miners behind reverse proxies.

_Proposed change_
Add with_rpc_port(u16) and with_web_port(u16) builder methods to MinerFactory:
```
let factory = MinerFactory::new()
    .with_rpc_port(14028)
    .with_web_port(8080);

let miner = factory.get_miner(ip).await?;
```
The ports propagate through discovery probes and data-collection backends. This requires a breaking change to FirmwareEntry::build_miner (trait signature) and the two utility functions.

nkatha23:
Fixes https://github.com/256foundation/asic-rs/issues/370

Replaces `Vec<MinerCommand>` in `get_discovery_commands()` with`Vec<DiscoveryCommand>`, a new type that carries an explicit `port: u16` alongside each command. Discovery probes now fire on the port declared by each firmware entry rather than hardcoded constants in `util.rs`.
Existing `send_rpc_command` / `send_web_command` signatures are unchanged and  backend data-collection call sites as well.


asic-rs-core/src/data/command.rs
```diff
@@ -3,6 +3,14 @@ use serde_json::Value;
 
 use crate::errors::RPCError;
 
+/// Command sent during miner discovery. Carries an explicit port so non-standard
+/// deployments can probe on remapped ports without changing global defaults.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum DiscoveryCommand {
+    Rpc { command: &'static str, port: u16 },
```
b-rowan: Would like to leave this as `RPC` in caps.

b-rowan (changes requested):
Just a small style change.  Looks good other than that.

b-rowan (approved):

